PR #1, opened 2022-03-22
Add hacky support for multi-line TSV entry parsing

lazaroclapp:
While, long term, we need a way for NullAway's serialization
logic to escape characters like tabs and new lines when producing
TSV data (or else quote the items as per RFC 4180), this change
allows us to successfully parse `errors.tsv` when the `MESSAGE`
field includes potentially multi-line code fragments.

This relies on two facts specific to `errors.tsv`:

* Only the second element in the TSV can ever contain tabs or new lines
* If this element contains new lines, they are immediately followed by
  whitespace due to code indentation conventions (either spaces or
  tabs).

Based on this, when reading the TSV files, we consume and concatenate
lines together until we find the next line that does NOT start with
whitespace. Then, we split the resulting text by TAB characters, and,
if we are in the multi-line case, and the number of TABs isn't exactly
4, then we concatenate parst [1:-2] together before passing them to
the factory.

Note that this approach will break if:
- Any TSV other than `errors.tsv` has multi-line entries, particularly
  if it has multi-line entries at a position other than the second
  column or has more or less than four colums
- There is any NullAway error message that is both multi-line and has
  no indenting whitespace.

Internally, this seems to work so far, but is no substitute to a proper
fix on NullAway's serialization.

### DIFF
--- a/core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
@@ -59,26 +59,37 @@ public class Index<T extends Hashable> {
   public void index() {
     items.clear();
     try (BufferedReader br = new BufferedReader(new FileReader(this.path.toFile()))) {
-      String line;
+      String entry = "";
+      String nextLine;
       br.readLine();
-      while ((line = br.readLine()) != null) {
-        if (line.contains("'")) {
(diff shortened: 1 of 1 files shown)


nimakarimipour:
Thank you very much @lazaroclapp. This looks good to me. I will work on the fix on the NullAway serialization as well.

PR #3, opened 2022-03-29
Additional fixes to multi-line TSV entry parsing

lazaroclapp:
This is a follow up to PR #1 with extra fixes to our TSV entry parsing
logic:

1) We resolve an off-by-one error introduced in PR #1 where the last
   line of a TSV file would be skipped (since the multi-line support
   read ahead for `nextLine` but would stop processing entries as soon
   as `nextLine = null`. We have fixed this here so that the loop stops
   only after `entry` is null)
2) We account for multi-line entries that might have empty lines
   (`nextLine.equals("")`) which don't technically start with
   whitespace.
3) We reintroduce line breaks into the message field of an `Error`
   object if those line breaks are present in the `MESSAGE` column
   of the TSV.
4) We fix incorrect indexing in `tsvValues`.

### DIFF
--- a/core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
@@ -24,9 +24,11 @@
 
 package edu.ucr.cs.riple.core.metadata.index;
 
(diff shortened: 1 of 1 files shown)


nimakarimipour:
@lazaroclapp Thank you very much for the fix, Looked good to me and merged it.

PR #7, opened 2022-06-08
Use shadow plugin for combined jar

msridhar:
WIP, edit as you see fit


nimakarimipour:
This is great. Thank you very much for the fix. I will merge it now.

PR #18, opened 2022-07-14
Fix #17

### LINKED ISSUE #17: Scanner Checker Fails when not activated.
**Describe the bug**
`Scanner` checker throws `null-pointer-exception` when the checker is not activated.

**To Reproduce**
1. Build the project when the config file is not passed as error prone flag.

**Expected behavior**
When the config file is not passed, the checker should not terminate with exception, instead it should silently quit from every method.

nimakarimipour:
Fixes #17.

### DIFF
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/DummyOptionsConfig.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/DummyOptionsConfig.java
@@ -27,33 +27,37 @@ package edu.ucr.cs.riple.scanner;
 import java.nio.file.Path;
 
 public class DummyOptionsConfig implements Config {
+
+  static final String ERROR_MESSAGE =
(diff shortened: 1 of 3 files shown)


lazaroclapp:
Took a look at this. I get how it works, but is it necessary to pass nulls this deeply into the code just to end up retrieving default values that produce a no-op analysis? Wouldn't it be easier to just set an `enabled` flag somewhere when passed the required configuration, and then check that flag at the start of each `Scanner.matchX` method and do nothing if it's `false`?

nimakarimipour:
@lazaroclapp as we discussed, I mistakenly assumed that the Checker constructor is called when the checker is `off`. The actual problem was that a different name space was used when sending error prone flags. The better way is to still throw an `Exception` if `ConfigPath` is `null`. Will reopen #17  now and update the code.

PR #19, opened 2022-07-14
Adds a guard for wrong configurations.

### LINKED ISSUE #17: Scanner Checker Fails when not activated.

nimakarimipour:
This PR is a follow up to GH-18 and resolves #17. With this PR, an exception will be thrown at the time a wrong configuration for the `Scanner` checker (`null` value for `Output` or `ConfigPath`) and will prevent propagation of `null` value through the program.

### DIFF
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
@@ -26,6 +26,7 @@ package edu.ucr.cs.riple.scanner;
 
 import com.google.common.base.Preconditions;
 import java.nio.file.Path;
+import javax.annotation.Nonnull;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
@@ -49,6 +50,7 @@ public interface Config {
 
   Serializer getSerializer();
 
+  @Nonnull
   Path getOutputDirectory();
 
   class Builder {
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
@@ -39,12 +39,12 @@ import org.xml.sax.SAXException;
 
 public class ErrorProneCLIFlagsConfig implements Config {
 
-  @Nonnull public final Path outputDirectory;
-  public final boolean methodTrackerIsActive;
-  public final boolean fieldTrackerIsActive;
-  public final boolean callTrackerIsActive;
-  public final boolean classTrackerIsActive;
-  public final Serializer serializer;
+  @Nonnull private final Path outputDirectory;
+  private final boolean methodTrackerIsActive;
+  private final boolean fieldTrackerIsActive;
+  private final boolean callTrackerIsActive;
+  private final boolean classTrackerIsActive;
+  private final Serializer serializer;
   static final String EP_FL_NAMESPACE = "Scanner";
   static final String FL_CONFIG_PATH = EP_FL_NAMESPACE + ":ConfigPath";
 
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
@@ -24,11 +24,13 @@
 
 package edu.ucr.cs.riple.scanner;
 
+import com.google.common.base.Preconditions;
 import com.google.errorprone.ErrorProneFlags;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import javax.annotation.Nonnull;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
@@ -37,7 +39,7 @@ import org.xml.sax.SAXException;
 
 public class ErrorProneCLIFlagsConfig implements Config {
 
-  public final Path outputDirectory;
+  @Nonnull public final Path outputDirectory;
   public final boolean methodTrackerIsActive;
   public final boolean fieldTrackerIsActive;
   public final boolean callTrackerIsActive;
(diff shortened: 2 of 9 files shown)


scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
```diff
@@ -49,6 +50,7 @@ public interface Config {
 
   Serializer getSerializer();
 
+  @Nonnull
```
lazaroclapp, line 53: Are we using a checker that looks at `@Nonnull`? Or the IDE? Wonder if it's worth the effort to get the scanner/annotator code to check with NullAway. Probably low priority right now, though, to be honest :)

(No action needed on this, just wondering)

scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
```diff
@@ -46,34 +48,26 @@ public class ErrorProneCLIFlagsConfig implements Config {
...
-        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
-        DocumentBuilder builder = factory.newDocumentBuilder();
-        document = builder.parse(Files.newInputStream(Paths.get(configFilePath)));
-        document.normalize();
-      } catch (IOException | SAXException | ParserConfigurationException e) {
-        throw new RuntimeException("Error in reading/parsing config at path: " + configFilePath, e);
-      }
+    Preconditions.checkNotNull(
+        configFilePath,
+        "Error in Scanner Checker configuration, should be set with via error prone flag: (-XepOpt:Scanner:ConfigPath)");
```
lazaroclapp, line 55: Do we have a test for this? We probably want a test case that shows that an error is thrown if `FL_CONFIG_PATH` is not passed.

scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
```diff
@@ -37,7 +39,7 @@
 
 public class ErrorProneCLIFlagsConfig implements Config {
 
-  public final Path outputDirectory;
+  @Nonnull public final Path outputDirectory;
```
lazaroclapp, line 42: Shouldn't all these fields be private? Particularly if we are calling the methods from `Config` consistently to get this info through those.

scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
lazaroclapp: If they are easy and not too time consuming right now, some other test cases for malformed config might make sense, but, at least we want the basic one of what happens if the flag is missing :)

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
nimakarimipour: Currently we are not using any checkers to look at these annotations. However, I think this is a really good idea and will add it eventually. #21 

I added this annotation just to make sure I am explicitly mentioning this property and also IDE will give some basic hints if we violate it.

lazaroclapp: Sounds good. Note we might want to remove them if/when we enroll this code in NullAway, given that non-null is the default. For now, though, they are definitely useful if the IDE is checking them, even in a less sound way.

scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
nimakarimipour: fe8992b6541de8c7128d8f621b399980c7a794e5

Added two guards, first the config path must be sent via error prone flag, and the config.xml file should contain a output dir value.

nimakarimipour:
@ketkarameya I added you to this project, I cannot ask for a request a review until you accept the invitation, let me know when you accept the invitation and I will add you as a reviewer for this PR :)

nimakarimipour:
@lazaroclapp This is ready for another round of review!

scanner/src/main/java/edu/ucr/cs/riple/scanner/ErrorProneCLIFlagsConfig.java
```diff
@@ -86,7 +80,7 @@ public ErrorProneCLIFlagsConfig(ErrorProneFlags flags) {
     this.classTrackerIsActive =
         XMLUtil.getValueFromAttribute(document, "/scanner/class", "active", Boolean.class)
             .orElse(false);
-    this.serializer = (this.outputDirectory == null) ? null : new Serializer(this);
+    this.serializer = new Serializer(this);
```
ketkarameya, line 83: I see that you removed the null check for `this.outputdirectory` but i cant see the check moved inside the serializer constructor

scanner/src/main/java/edu/ucr/cs/riple/scanner/Serializer.java
```diff
@@ -53,8 +53,8 @@ public class Serializer {
   public static final String METHOD_INFO_FILE_NAME = "method_info.tsv";
   public static final String CLASS_INFO_FILE_NAME = "class_info.tsv";
 
-  public Serializer(ErrorProneCLIFlagsConfig config) {
-    Path outputDirectory = config.outputDirectory;
+  public Serializer(Config config) {
+    Path outputDirectory = config.getOutputDirectory();
```
ketkarameya, line 57: https://github.com/nimakarimipour/NullAwayAnnotator/pull/19/files

scanner/src/test/java/edu/ucr/cs/riple/scanner/ConfigurationTest.java
```diff
@@ -0,0 +1,122 @@
...
+        Document doc = docBuilder.newDocument();
+        Element rootElement = doc.createElement("scanner");
+        doc.appendChild(rootElement);
+        TransformerFactory transformerFactory = TransformerFactory.newInstance();
+        Transformer transformer = transformerFactory.newTransformer();
+        DOMSource source = new DOMSource(doc);
+        StreamResult result = new StreamResult(config.toFile());
+        transformer.transform(source, result);
+      } catch (ParserConfigurationException | TransformerException e) {
+        throw new RuntimeException("Error happened in writing config.", e);
```
ketkarameya, line 104: in -> while

lazaroclapp:
Sorry for the last few messages not being associated with a line @nimakarimipour , forgot I was looking at a commit rather than the PR. You can see them in [context](https://github.com/nimakarimipour/NullAwayAnnotator/commit/fe8992b6541de8c7128d8f621b399980c7a794e5) here 😅

scanner/src/main/java/edu/ucr/cs/riple/scanner/Serializer.java
nimakarimipour: Sorry I didn't fully understand this comment. Would you please elaborate.

scanner/src/test/java/edu/ucr/cs/riple/scanner/ConfigurationTest.java
nimakarimipour: 42b84b5fc2f8224b9eb1f8fa29009aa3b579febd

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/SerializationTestHelper.java
```diff
@@ -86,17 +99,28 @@ public SerializationTestHelper<T> setOutputFileNameAndHeader(String fileName, St
...
     try {
-      Files.deleteIfExists(outputPath);
+      Files.deleteIfExists(outputFilePath);
     } catch (IOException ignored) {
-      throw new RuntimeException("Failed to delete older file at: " + outputPath);
+      throw new RuntimeException("Failed to delete older file at: " + outputFilePath);
     }
+  }
+
+  public void doTest(Class<? extends Exception> exception, String expectedErrorMessage) {
```
lazaroclapp, line 113: 👍 

Appreciate that we separated the two versions of this method. Looking at it now, can I suggest renaming this one `doTestExpectingError(...)`, just to make it even more obvious what is happening when reading the test cases.

lazaroclapp (approved):
Looks fine to land to me if @ketkarameya's comments have also been addressed. As an optional suggestion, other than the method rename, one thing I would recommend is for the new public methods of SerializationTestHelper to have javadoc comments.

nimakarimipour:
@ketkarameya This is ready for review, I think @lazaroclapp approved it. Please let me know once you approve this PR and I will land it, Thank you :)

ketkarameya (approved):
LGTM.

PR #22, opened 2022-07-18
Serialize method visibility / return type

nimakarimipour:
This PR adds visibility and return type of method to list of informations serialized on output.

```
id, class, method, parent, size, flags, nullable, parameters, uri
```

will be changed to:

```
id, class, method, parent, size, flags, nullable, visibility, non-primitive-return, parameters, uri
```

If the method is returning a `primitive` type of `void` the `non-primitive-return` value will be set to `false` and `true` otherwise.

### DIFF
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
@@ -109,6 +109,7 @@ public class MethodInfo {
         Arrays.toString(annotFlags),
         String.valueOf(hasNullableAnnotation),
         getVisibilityOfMethod(),
+        String.valueOf(!symbol.getReturnType().isPrimitiveOrVoid()),
         Arrays.toString(parameterNames),
         uri.getPath());
   }
@@ -129,6 +130,8 @@ public class MethodInfo {
         + "nullable"
         + "\t"
         + "visibility"
+        + "\t"
+        + "non-primitive-return"
         + "\t"
         + "parameters"
         + "\t"
@@ -170,8 +173,8 @@ public class MethodInfo {
   /**
    * Return string value of visibility
    *
-   * @return "public" if public, "private" if private and "package" if the method has package
-   *     visibility.
+   * @return "public" if public, "private" if private, "protected" if protected and "package" if the
+   *     method has package visibility.
    */
   private String getVisibilityOfMethod() {
     Set<Modifier> modifiers = symbol.getModifiers();
@@ -180,6 +183,9 @@ public class MethodInfo {
     }
     if (modifiers.contains(Modifier.PRIVATE)) {
       return "private";
+    }
+    if (modifiers.contains(Modifier.PROTECTED)) {
+      return "protected";
     }
     return "package";
   }
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
@@ -39,7 +39,8 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import org.checkerframework.checker.nullness.qual.Nullable;
+import javax.annotation.Nullable;
+import javax.lang.model.element.Modifier;
 
 public class MethodInfo {
   public final Symbol.MethodSymbol symbol;
@@ -112,6 +113,8 @@ public class MethodInfo {
         + "\t"
         + hasNullableAnnotation
         + "\t"
+        + getVisibilityOfMethod()
+        + "\t"
         + Arrays.toString(parameterNames)
         + "\t"
         + uri.getPath();
@@ -131,6 +134,8 @@ public class MethodInfo {
         + "flags"
         + "\t"
         + "nullable"
+        + "\t"
+        + "visibility"
         + "\t"
         + "parameters"
         + "\t"
@@ -167,5 +172,22 @@ public class MethodInfo {
                   }
                 })
             .toArray(String[]::new);
+  }
+
+  /**
+   * Return string value of visibility
+   *
+   * @return "public" if public, "private" if private and "package" if the method has package
+   *     visibility.
+   */
+  private String getVisibilityOfMethod() {
+    Set<Modifier> modifiers = symbol.getModifiers();
+    if (modifiers.contains(Modifier.PUBLIC)) {
+      return "public";
+    }
+    if (modifiers.contains(Modifier.PRIVATE)) {
+      return "private";
+    }
+    return "package";
   }
 }
--- a/scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
+++ b/scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
@@ -99,25 +99,18 @@ public class MethodInfo {
   @Override
   public String toString() {
     Preconditions.checkArgument(symbol != null, "Should not be null at this point.");
-    return id
-        + "\t"
-        + (clazz != null ? clazz.flatName() : "null")
-        + "\t"
-        + symbol
-        + "\t"
-        + parent
-        + "\t"
-        + symbol.getParameters().size()
-        + "\t"
-        + Arrays.toString(annotFlags)
-        + "\t"
-        + hasNullableAnnotation
-        + "\t"
-        + getVisibilityOfMethod()
-        + "\t"
-        + Arrays.toString(parameterNames)
-        + "\t"
-        + uri.getPath();
+    return String.join(
+        "\t",
+        String.valueOf(id),
+        (clazz != null ? clazz.flatName() : "null"),
+        symbol.toString(),
+        String.valueOf(parent),
+        String.valueOf(symbol.getParameters().size()),
+        Arrays.toString(annotFlags),
+        String.valueOf(hasNullableAnnotation),
+        getVisibilityOfMethod(),
+        Arrays.toString(parameterNames),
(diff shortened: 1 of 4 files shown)


scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
```diff
@@ -112,6 +113,8 @@ public String toString() {
         + "\t"
```
ketkarameya, line 113: Do we want to use String.format() here? I think its more efficient than concatenation of literals .

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
```diff
@@ -168,4 +173,21 @@ public void setParameterNames(List<? extends VariableTree> parameters) {
...
             .toArray(String[]::new);
   }
+
+  /**
+   * Return string value of visibility
+   *
+   * @return "public" if public, "private" if private and "package" if the method has package
+   *     visibility.
+   */
+  private String getVisibilityOfMethod() {
```
ketkarameya, line 183: Do we need to reason about the "default" modifier too?

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/MethodInfoDisplay.java
```diff
@@ -104,6 +107,9 @@ public String toString() {
         + ", hasNullableAnnotation='"
         + hasNullableAnnotation
         + '\''
```
ketkarameya, line 109: nit: Same as above. String.format() over concatenation.

scanner/src/test/java/edu/ucr/cs/riple/scanner/MethodInfoTest.java
```diff
@@ -36,13 +36,13 @@ public class MethodInfoTest extends ScannerBaseTest<MethodInfoDisplay> {
 
   private static final DisplayFactory<MethodInfoDisplay> METHOD_DISPLAY_FACTORY =
       values -> {
-        Preconditions.checkArgument(values.length == 9, "Expected to find 9 values on each line");
+        Preconditions.checkArgument(values.length == 10, "Expected to find 9 values on each line");
```
ketkarameya, line 39: 10 values?

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
nimakarimipour: We are not reasoning about the `default` modifier, however I believe it is better to serialize full information rather than just the information we currently need to avoid further PRs :)

There are two alternatives to this approach:
1. write `public` for `public method` and `private` for `private method` and `null` for package visibility.
2. only write `true` for public methods and `false` for all others.

I prefer the approach we are already taking as it is more readable and understandable. Let me know what you think of this. Also I am happy to hear your thoughts @lazaroclapp as well :)

scanner/src/test/java/edu/ucr/cs/riple/scanner/MethodInfoTest.java
nimakarimipour: Yes, `visibility` is added now.

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/MethodInfoDisplay.java
```diff
@@ -46,6 +47,7 @@ public MethodInfoDisplay(
       String size,
       String flags,
       String hasNullableAnnotation,
+      String visibility,
```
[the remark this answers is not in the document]
nimakarimipour: This `display` classes are just instances to help us read serialized outputs, therefore they need contain all the serialized information on the output including the `visibility` to run the tests successfully.

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
ketkarameya: I see. I prefer the current strategy over 2.

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
nimakarimipour: I am not sure if using `String.format()` can lead to a measurable speed up here. Generally speaking with the current structure of the `core` and `scanner` module, with our experiments `98-99 %` of the time in the process is spent in re-building the target project :) Please let me know what you think of this and still interested in converting the concatenations to `String.format()`.

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/MethodInfoDisplay.java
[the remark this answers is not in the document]
ketkarameya: Maybe my suggestion didn't come across. What I was suggesting is `MethodInfoDisplay(String ... args)` because in method info test, you have to awkwardly parse the array. But it's fine, it's just a nit.  Since I cant actually comment upon the functionality, I am acting as a human linter. :P

ketkarameya:
The changes LGTM. I think that the branch protection rule "does not require approval" . should this be changed? Idk if this is on purpose.

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
lazaroclapp: Besides performance, `String.format()` can sometimes help with readability. However, in this particular case, I'd argue it will make this harder, not easier, to read and update this list (i.e. counting placeholders in the template string). I would consider `return String.join("\t", ...)` here, though, keeping the elements each on their own line. That would hopefully have a fast implementation of concatenation underneath, but also, is arguably more readable than what we currently have. 

Either way, I am fine with what we have in this case, just giving options :)

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
lazaroclapp: The current approach makes sense to me, but I believe we are missing a check for `Modifier.PROTECTED`.

scanner/src/test/java/edu/ucr/cs/riple/scanner/MethodInfoTest.java
```diff
@@ -90,6 +92,61 @@ public void BasicTest() {
...
                 "[]",
                 "false",
+                "public",
+                "[]",
+                "edu/ucr/A.java"))
+        .doTest();
+  }
+
+  @Test
+  public void visibilityTest() {
```
lazaroclapp, line 102: I can think of another three worthwhile cases to test:

1. Protected methods
2. Methods with modifiers unrelated to visibility (such as `default`, `abstract`, `static`, etc) in addition to visibility modifiers. No need to test all combinations, but it's worth having an example that has more than a single modifier on a method signature.
3. Methods of an interface, rather than a class. Does this approach figure out that "public" is [implicit](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.4) for interface methods? I think it might. I think that information may already be explicit at the compiler level, but we should check.

My personal instinct would be to add (1) to the test case below, and make new separate test cases for (2) and (3).

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/MethodInfoDisplay.java
lazaroclapp: This one actually does feel more to me like a case where `String.format()` could make the code more readable than plain concatenation.

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
nimakarimipour: I did not know about `String.join()`, thanks for noting it.

753602ebbcd8dc01c2b38b517532cf9eabcf0ac3

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/MethodInfoDisplay.java
nimakarimipour: These are generated by Intellij and the only purpose is to just show a hint to the user what wrong output failed the test. Is it really worth it to undo a generated code here :) ?

scanner/src/test/java/edu/ucr/cs/riple/scanner/MethodInfoTest.java
nimakarimipour: I added the method return type as well. Just a flag showing the return type is non-primitive. In the upstream analysis we should not waste builds with methods which are return `void/primitive`. Updated this PR description as well.

edccb99ef8288dfcb90cbc155a727986340f7998

nimakarimipour:
@lazaroclapp Just a quick note, I am going to remove the serialization parameter names soon since they are needed anymore. Please find the issue at #23.

scanner/src/test/java/edu/ucr/cs/riple/scanner/tools/MethodInfoDisplay.java
lazaroclapp: Fair enough. Fine to keep the generated version, then.

ketkarameya: lol! This stirred up too much conversation :O

lazaroclapp (approved):
LGTM

PR #29, opened 2022-07-21
Prevent crash when a target class is not located with better logging mechanism.

### LINKED ISSUE #27: Annotator fails when a Fix targeting an element in a generated code is suggested.
**Describe the bug**
In the process of measuring effectiveness of fixes, if a fix is targeting an element in the generated code, `Injector` will throw an exception that it could not find the class which does not exist.

**To Reproduce**
Use a `lombok` builder and target a method in the `builder` class.

**Expected behavior**
`Injector` should report these classes silently when the class is not found and should not terminate the process.

**Screenshots**
Not-Applicable.

**Additional context**
No Additional Context.

nimakarimipour:
This PR fixes #27 by adding a new type of exception `TargetClassNotFound`. With this PR annotator will not crash if the class is not found, and will instead log this information for later debugging. 

Also added a test to verify, annotator will not fail and the desired error message is correctly reported in the `err` console.

### DIFF
--- a/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
+++ b/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
@@ -108,7 +108,7 @@ public class Helper {
   }
 
   /**
-   * Extracts the callable simple name from callable signature. (e.g. in input "run(Object i)"
+   * Extracts the callable simple name from callable signature. (e.g. on input "run(Object i)"
    * returns "run").
    *
    * @param signature callable signature in string.
--- a/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
+++ b/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
@@ -229,10 +229,7 @@ public class Helper {
   private static Node findInnerClass(Node cursor, String name, int index)
       throws TargetClassNotFound {
     final List<Node> candidates = new ArrayList<>();
-    walk(
-        cursor,
-        candidates,
-        node -> isDeclarationWithName(node, name));
+    walk(cursor, candidates, node -> isDeclarationWithName(node, name));
     if (index >= candidates.size()) {
       throw new TargetClassNotFound("Non-Direct-Inner-Class", index + name, cursor);
     }
--- a/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
+++ b/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
@@ -232,16 +232,7 @@ public class Helper {
     walk(
         cursor,
         candidates,
-        node -> {
-          Optional<Node> parent = node.getParentNode();
-          if (parent.isEmpty()) {
-            return false;
-          }
-          if (parent.get().equals(cursor)) {
-            return false;
-          }
-          return isDeclarationWithName(node, name);
-        });
+        node -> isDeclarationWithName(node, name));
     if (index >= candidates.size()) {
       throw new TargetClassNotFound("Non-Direct-Inner-Class", index + name, cursor);
     }
--- a/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
+++ b/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
@@ -29,10 +29,8 @@ import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.PackageDeclaration;
 import com.github.javaparser.ast.body.AnnotationDeclaration;
 import com.github.javaparser.ast.body.BodyDeclaration;
-import com.github.javaparser.ast.body.CallableDeclaration;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.EnumDeclaration;
-import com.github.javaparser.ast.body.Parameter;
 import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.google.common.base.Preconditions;
@@ -52,60 +50,6 @@ import me.tongfei.progressbar.ProgressBarStyle;
 
 /** A utility class. */
 public class Helper {
-
-  /** Utility class to match {@link CallableDeclaration} with their signatures in {@code String}. */
-  public static class SignatureMatcher {
-
-    /** Simple name of the callable. */
-    private final String callableName;
-    /** List of parameters detected from signature in string. */
-    private final List<String> parameterTypes;
-
-    /**
-     * Constructor to make a matcher instance.
-     *
-     * @param signature Signature to process.
-     */
-    public SignatureMatcher(String signature) {
-      this.callableName = extractCallableName(signature);
-      this.parameterTypes = extractParamTypesOfCallableInString(signature);
-    }
-
-    /**
-     * Checks if callableDec's signature is equals to parsed signature.
-     *
-     * @param callableDec callable declaration node.
-     * @return true, if signature matches the callable and false otherwise.
-     */
-    public boolean matchesCallableDeclaration(CallableDeclaration<?> callableDec) {
-      // match callable names.
-      if (!callableDec.getName().toString().equals(callableName)) {
-        return false;
-      }
-      // match parameter types.
-      List<String> paramTypesFromCallableSignature =
-          extractParamTypesOfCallableInString(callableDec);
-      if (parameterTypes.size() != paramTypesFromCallableSignature.size()) {
-        return false;
-      }
-      int size = parameterTypes.size();
-      for (int i = 0; i < size; i++) {
-        String callableType = parameterTypes.get(i);
-        String signatureType = paramTypesFromCallableSignature.get(i);
-        if (signatureType.equals(callableType)) {
-          continue;
-        }
-        String simpleCallableType = simpleName(callableType);
-        String simpleSignatureType = simpleName(signatureType);
-        if (simpleCallableType.equals(simpleSignatureType)) {
-          continue;
-        }
-        // Param types are different.
-        return false;
-      }
-      return true;
-    }
-  }
 
   /**
    * Extracts the callable simple name from callable signature. (e.g. on input "run(Object i)"
@@ -134,27 +78,6 @@ public class Helper {
       }
     }
     return ans.toString();
-  }
-
-  /**
-   * Extracts parameters type from a {@link CallableDeclaration}.
-   *
-   * @param callableDec callable declaration instance.
-   * @return List of parameters type in string.
-   */
-  private static List<String> extractParamTypesOfCallableInString(
-      CallableDeclaration<?> callableDec) {
-    ArrayList<String> paramTypes = new ArrayList<>();
-    for (Parameter param : callableDec.getParameters()) {
-      if (param != null) {
-        String typeInString = param.getType().asString();
-        if (param.isVarArgs()) {
-          typeInString += "...";
-        }
-        paramTypes.add(typeInString);
-      }
-    }
-    return paramTypes;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -210,6 +219,9 @@ private static Node findTopLevelAnonymousClass(Node cursor, int index) {
           }
           return false;
         });
+    if (index >= candidates.size()) {
+      throw new TargetClassNotFound("Top-Level-Anonymous-Class", "$" + index, cursor);
+    }
```
lazaroclapp, line 224: This might not be quite as easy as this, though... what happens if there are `Foo$1`, `Foo$2`, and `Foo$3`, but it is `Foo$2` that is generated? We might some additional information from NullAway to handle this properly in the case of anonymous classes specifically (Fortunately, Lombok generated _anonymous_ classes is not something I have run into, at least outside of Lombok-generated methods. Still, I wonder if we should fail loudly in that case, rather than silently pick the wrong anonymous inner class)

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -149,10 +148,14 @@ private static TypeDeclaration<?> findTopLevelClassDeclarationOnCompilationUnit(
...
     Optional<ClassOrInterfaceDeclaration> interfaceDeclaration = tree.getInterfaceByName(name);
-    return interfaceDeclaration.orElseGet(() -> tree.getClassByName(name).orElse(null));
+    if (interfaceDeclaration.isPresent()) {
+      return interfaceDeclaration.get();
+    }
+    throw new TargetClassNotFound("Top-Level", name, tree);
   }
 
-  public static Node findTopLevelDirectInnerClass(Node cursor, String name) {
+  private static Node findTopLevelDirectInnerClass(Node cursor, String name)
```
lazaroclapp, line 157: Even though these are private methods, I think they would benefit from Javadocs. What is the difference between a "Top level direct inner class" and a "top level inner class", for example?

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -161,26 +164,31 @@ public static Node findTopLevelDirectInnerClass(Node cursor, String name) {
...
     walk(
         cursor,
         candidates,
         node -> {
           Optional<Node> parent = node.getParentNode();
-          if (!parent.isPresent()) {
+          if (parent.isEmpty()) {
             return false;
           }
           if (parent.get().equals(cursor)) {
```
lazaroclapp, line 184: Question: What is this line checking for?

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
lazaroclapp: Either way: worth tracking as a future task somewhere. Not high priority, but definitely a potential pitfall/bug.

lazaroclapp (commented):
Some comments. Also, a slight meta-comment: it's worth it for the PR to have a title that directly explains what the fix is about rather than an indirect reference to an issue. You can always link the issue in the description, but the commit message is a) more easily visible and b) more permanent in the codebase than any issue on any tracker.

injector/src/main/java/edu/ucr/cs/riple/injector/exceptions/TargetClassNotFound.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package edu.ucr.cs.riple.injector.exceptions;
+
+import com.github.javaparser.ast.Node;
+
+public class TargetClassNotFound extends Exception {
+
+  static final String REPORT_MESSAGE =
+      "If you do see the class on the source code, please report this bug at: https://github.com/nimakarimipour/NullAwayAnnotator/issues. Thank you!";
```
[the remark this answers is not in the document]
nimakarimipour: 67effbd9328f8c587712d6ceb5b8d010bc9ab0fc

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: Actually this was needed for previous approach, with the new `walk` method I wrote, this is not required anymore. Deleted it. 92b769bb4cd4e901a774a89f74e3d690f9cc40d6

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: Sure, that's a good point. Will create an issue once this lands, to track it.

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
```diff
@@ -468,4 +475,32 @@ public void class_search_benchmark_22() {
                 true))
         .start(true);
   }
+
+  @Test
+  public void proper_report_for_class_not_found_test() {
```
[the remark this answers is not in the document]
nimakarimipour: Actually I wanted to divide it into two steps, but that requires hardcoding the expected error message. Which, if we change the message, we need to update the tests as well. However, it makes more sense to divide it to two cases in favor of cohesiveness of each test.

9d5f3188f1500967ef879e7979ee0cc87b751adc

Please let me know if you prefer the older version.

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: Added javadocs for all methods in this class. I am planning to add javadoc through every PR :) c623aacf3634525f48470ead8e8d3506348029d6

Direct inner class requires the class to be connected to parent directly, and the other one does not have this constraint. `top level` name is also removed from them, they are on `findDirectInnerClass` and `findInnerClass`.

nimakarimipour:
@lazaroclapp This is ready for another review.

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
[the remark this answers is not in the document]
lazaroclapp: Hardcoding the expected message in tests and having to change the tests when behavior changes is very reasonable. In production code, I'd err on the side of not hardcoding things and avoiding repetition, since that makes the code fragile. But within unit tests, clarity is more important, and I'd argue that failing when existing behavior changes is a feature, not a bug, of a comprehensive test suite :)

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -50,8 +50,16 @@
 import me.tongfei.progressbar.ProgressBar;
 import me.tongfei.progressbar.ProgressBarStyle;
 
+/** A utility class. */
 public class Helper {
 
+  /**
+   * Extracts the callable simple name from callable signature. (e.g. in input "run(Object i)"
```
lazaroclapp, line 57: "on input"

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
```diff
@@ -493,14 +503,32 @@ public void proper_report_for_class_not_found_test() {
                 "javax.annotation.Nullable",
                 true))
         .start();
-    String errOutput = err.toString();
+    assert err.toString().equals(expectedErrorMessage);
```
lazaroclapp, line 506: Don't use java assert directly in tests, prefer JUnit's assertion APIs, such as `assertEquals(...)` or `assertTrue(...)`. That way we don't depend on whether assertions are kept by the compiler or not (subject to compiler configuration)

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
```diff
@@ -477,7 +477,17 @@ public void class_search_benchmark_22() {
   }
 
   @Test
-  public void proper_report_for_class_not_found_test() {
+  public void proper_report_in_err_std_for_class_not_found_test() {
+    String expectedErrorMessage =
```
lazaroclapp, line 481: Thinking about these long errors, we probably should just be testing a substring. Something like:

```
String expectedErrorMessage = "Could not find class of type: Top-Level with name: NotIncluded"
[...]
assertTrue(err.toString().contains(expectedErrorMessage));
```

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -131,6 +163,16 @@ private static void walk(Node cursor, List<Node> candidates, Predicate<Node> pre
         });
   }
 
+  /**
+   * Finds Top-Level type declaration within a {@link CompilationUnit} tree by name, this node is
+   * direct child of the compilation unit tree and can be: [{@link ClassOrInterfaceDeclaration},
```
lazaroclapp, line 168: Missing article? Either "a direct child" or "the direct child"

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -66,43 +74,57 @@ public static String extractCallableName(String signature) {
...
   }
 
+  /**
+   * Checks if callableDec's signature is equals to signature passed in the argument.
+   *
+   * @param callableDec callable declaration node.
+   * @param signature signature of the callable in string.
+   * @return true, if signature matches the callable and false otherwise.
+   */
   public static boolean matchesCallableSignature(
```
lazaroclapp, line 92: Definitely not for this PR, but I notice there are a lot of methods here comparing signatures in String format to `CallableDeclaration` by parsing the string multiple times and extracting different portions of it: wouldn't it be more useful to have a single function parsing a string into it's corresponding `CallableDeclaration`? In fact, if the only comparison we need is strict equality, a well crafted `CallableDeclaration.toString()` method could be all we need here, no parsing required.

nimakarimipour: Hi @lazaroclapp, sure that sounds good. However, at the moment I was writing `injector` it wasn't merged into annotator as a sub-module and I was intending to make a robust tool that works with any given signature, not necessarily matching how `error prone` serializes methods signature. It now works with any given input as long as the types and names and correct (simple name or fully qualified name). However, what you mention about parsing it multiple times is a really good point and it worths to address it in this PR. Please find it here where we only parse each signature once. 13ee6712184d74d1edea5e155f3967f5b2c86b57.

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: 6079a72c488c2b651b698a6d88a0c7d29aaa784e

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
nimakarimipour: e8040841a6b3d56eb5d1a8798052acabcd760a11

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
nimakarimipour: e8040841a6b3d56eb5d1a8798052acabcd760a11

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: c5d0438d0fdadee93ad046724bab696703a88919

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -53,8 +53,62 @@
...
+    private final List<String> parameterTypes;
+
+    /**
+     * Constructor to make a matcher instance.
+     *
+     * @param signature Signature to process.
+     */
+    public SignatureMatcher(String signature) {
+      this.callableName = extractCallableName(signature);
+      this.parameterTypes = extractParamTypesOfCallableInString(signature);
```
lazaroclapp, line 71: Are `extractCallableName` and `extractParamTypesOfCallableInString` used outside this class? If not, let's make them private methods of `SignatureMatcher` and consider extracting this class as a top level class in its own file.

lazaroclapp (commented):
Mostly looks ready to me, but wondering if we can extract `SignatureMatcher` and its related methods into a top-level class. See below.

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: No they can be internal. Moved the Signature Matcher class to a top level class. Please find it here: c33e8b35e3db38385d303aacf237a2d5b7282677

lazaroclapp (approved):
LGTM! 🎉

nimakarimipour:
@ketkarameya Please let me know once you approve this and I will land this.

ketkarameya (approved):
LGTM !

PR #33, opened 2022-08-01
Add parallel scanning of downstream dependencies.

nimakarimipour:
Following GH-26 which enables Annotator to perform downstream dependency analysis, it requires running NullAway on modules multiple times. Since running NullAway on each module is completely independent of one another, in an environment where multiple execution cores are available, we can perform the analysis on all modules in parallel. This PR enables Annotator to benefit from environments where multiple cores are available.

#### Changes
1. It accepts a `tsv` file containing paths to `NullAway` and `Scanner` for each downstream dependency
2. It accepts a universal build command which can build all downstream dependencies in parallel.
3. For each downstream dependency a unique `output` directory is created which contains all its corresponding serialized data by checkers.
4. All loading containers are now loading their data cumulatively from from multiple sources.

#### Updates in flags
1. `-ddbc` or `--downstream-dependencies-build-command`: Command to build all downstream dependencies at once, this command must include changing directory from root to the target project
2. `-nlmlp` or `--nullaway-library-model-loader-path`: NullAway Library Model loader path
3. `-ddcp` or `--downstream-dependencies-config-path`: Path to tsv file containing path to nullaway and scanner config files.

#### Updates in JSON
```json
{
   "DOWNSTREAM_DEPENDENCY_ANALYSIS":
    {
         "ACTIVATION":
         "BUILD_COMMAND":
         "LIBRARY_MODEL_LOADER_PATH":
         "CONFIG_PATHS":
                [
                      {
                             "NULLAWAY":
                             "SCANNER":
                      }
                ]

    }
}
```
`ACTIVATION`: Activation flag for downstream dependency analysis.
`BUILD_COMMAND`: Command to build all downstream dependencies at once, this command must include changing directory from root to the target project
`LIBRARY_MODEL_LOADER_PATH`:  NullAway Library Model loader path
`CONFIG_PATHS`: Array of json object which includes path to nullaway and scanner config files with `NULLAWAY` and `SCANNER` keys


core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
```diff
@@ -82,26 +83,29 @@ public Index(Path path, Index.Type type, Factory<T> factory) {
...
-      }
-    } catch (IOException e) {
-      e.printStackTrace();
-    }
+    paths.forEach(
+        path -> {
+          try (BufferedReader br = Files.newBufferedReader(path, UTF_8)) {
+            String line = br.readLine();
+            if (line != null) {
+              line = br.readLine();
```
lazaroclapp, line 91: Please add a comment about what's happening here. Presumably this skipping the TSV headers, right?

lazaroclapp (commented):
Overall looks good to me, minor comments about checker suppressions and other nits. More importantly, do we have any tests for these new command line flags? Whether as part of this PR or a follow up? (I know the python script that passes these new flags internally is not ready yet, so I am wondering how we are testing this, even manually)

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Bank.java
```diff
@@ -50,18 +52,22 @@
   /** Factory instance to parse outputs. */
   private final Factory<T> factory;
   /** Path to file where outputs are stored. */
-  private final Path path;
+  private final ImmutableSet<Path> paths;
 
-  public Bank(Path path, Factory<T> factory) {
+  public Bank(Stream<Path> paths, Factory<T> factory) {
```
ketkarameya, line 57: I personally find it uncomfortable to see `Stream` being passed around because they are on-time-use only. Also based on the current usages, it would be cleaner to accept a `ImmuableSet`.

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
```diff
@@ -65,15 +66,15 @@ public enum Type {
   }
 
   /**
-   * Creates an instance of Index.
+   * Creates an instance of Index. Contents are accumulated from multiple sources.
    *
-   * @param path Path to the file to load the data from.
+   * @param paths ImmutableSet of paths to load the data from.
```
ketkarameya, line 71: Can u add some description around, what is the format expected for this file? maybe a sample row?

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
```diff
@@ -82,26 +83,29 @@ public Index(Path path, Index.Type type, Factory<T> factory) {
...
+    paths.forEach(
+        path -> {
+          try (BufferedReader br = Files.newBufferedReader(path, UTF_8)) {
+            String line = br.readLine();
+            if (line != null) {
+              line = br.readLine();
+            }
+            while (line != null) {
+              T item = factory.build(line.split("\t"));
+              total++;
```
ketkarameya, line 95: Make `total` a private field. And add a getter.  You don't want to be accidentally updating the field anywhere :)

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
```diff
@@ -82,26 +83,29 @@ public Index(Path path, Index.Type type, Factory<T> factory) {
...
-      e.printStackTrace();
-    }
+    paths.forEach(
+        path -> {
+          try (BufferedReader br = Files.newBufferedReader(path, UTF_8)) {
+            String line = br.readLine();
+            if (line != null) {
+              line = br.readLine();
+            }
+            while (line != null) {
```
ketkarameya, line 93: A "base case" check here for an empty line?

ketkarameya (approved):
Left some comments. Otherwise, it LGTM.

core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -458,6 +513,7 @@ public static class Builder {
     public String nullawayLibraryModelLoaderPath;
     public int depth = 1;
 
+    @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
nimakarimipour: Actually this method changed later (at the time of making unit tests), I removed some flags to make configuring easier. For instance we had two separate flags `--nullaway-config-paths` and `scanner-config-paths` and with downstream dependencies analysis activated, a path to a file containing these paths for the rest of the modules. I removed `--nullaway-config-paths` and `scanner-config-paths` flags and now we only require a path to a file which the first entry is for the target module and others are downstream dependencies.

There was always a warning when compiling `core` module and thought since I am modifying this method, add a suppress warning since it's just because of json creation.

core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -192,22 +235,33 @@ public Config(String[] args) {
     options.addOption(scannerConfigPathOption);
 
     // Down stream analysis
+    OptionGroup downstreamDepAnalysisOption = new OptionGroup();
```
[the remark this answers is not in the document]
nimakarimipour: @lazaroclapp Thank you very much for noticing this, I definitely did not mean that. I am making unit tests with the json option not the `cli` option. I fixed what I meant there. What I meant is that either all or none should all exist. As I was debugging rest of the flags, I realized due to changes in this PR some other bugs are in the `cli` flag configuration. I ended up making a whole new test class for configuration tests. Thanks again for noticing this. 43796acb24b5243bfb6748f25e76aa35a125617b

core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -387,8 +441,9 @@ private <T> OrElse<T> getValueFromKey(JSONObject json, String key, Class<T> klas
     }
   }
 
+  @SuppressWarnings({"SameParameterValue", "unchecked"})
```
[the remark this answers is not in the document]
nimakarimipour: First one is IntelliJ warning that the input of this function is always and suggests to inline it, the other one is regarding calling map on stream of object. We expect JSONArray to return List of JSON Object here.

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
nimakarimipour: f490c9278abb1fd31da6b8d56bd12aad4bb80e0a

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Bank.java
nimakarimipour: Thank you very much for noting that, I mistakenly assumed since they don't have `add` method, they are immutable. This also has some other problems as you mentioned. 706face8159aaeddaee2861e298432f9930aeacd

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
nimakarimipour: Sure, actually there is no expected header. These classes can index any type of `.tsv` file where they contain information for creating instances of `Enclosed`. Yes I do agree that javadoc was not clear enough, please find the updated version c93236ba89e4e746392620f1011b0a23327105d9 and let me know if I should add anything else. Thanks.

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
nimakarimipour: 4695aa7091841e051b1d0704fd960201d6457807

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
nimakarimipour: It is skipping the header of the TSV file. I added a comment on top of that. f490c9278abb1fd31da6b8d56bd12aad4bb80e0a

nimakarimipour:
@ketkarameya @lazaroclapp Thank you for your comments I resolved them all. This is ready for another review.

nimakarimipour:
FYI, added #34 that down stream dependency test are not included in the CI yet, will include it once `NullAway:0.9.9` is released.

lazaroclapp (approved):
Some minor nits, but, overall, this LGTM!

core/src/main/java/edu/ucr/cs/riple/core/Config.java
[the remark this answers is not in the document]
nimakarimipour: The input value for now is always: `CONFIG_PATHS`. I think as this the tool evolves configurations can be more complex and this will be called with different values in future. Pleas let me know if you think I should inline this for now and update it future :)

library-model-loader/build.gradle
```diff
@@ -31,7 +31,7 @@ archivesBaseName = "librarymodel"
 dependencies {
     compileOnly deps.apt.autoService
     annotationProcessor deps.apt.autoService
-    compileOnly deps.build.nullaway
+    implementation deps.build.nullaway
```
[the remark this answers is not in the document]
nimakarimipour: Yes that make sense. d36f7df5200af2b7071e5e34193cd2cf960f6a42

core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java
```diff
@@ -24,9 +24,37 @@
...
   public DownstreamAnalysisTest() {
-    super("downstream-dependency-test");
+    super("downstream-dependency-test", List.of("Target", "DepA", "DepB", "DepC"));
+  }
+
+  @Test
+  public void public_method_with_downstream_dependency_enabled() {
+    coreTestHelper
+        .addExpectedReports(
+            new TReport(new OnMethod("Foo.java", "test.target.Foo", "returnNullableBad(int)"), 5),
```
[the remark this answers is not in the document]
nimakarimipour: Added the comment: 02523bdd8c11fb16ceef536bcccac8c576b2b884

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
```diff
@@ -0,0 +1,192 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Nima Karimipour
```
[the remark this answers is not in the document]
nimakarimipour: Yes :))) Updated headers, all were wrong. Thanks for noticing this :)

PR #36, opened 2022-08-04
Rename Scanner module to TypeAnnotatorScanner.

### LINKED ISSUE #25: Rename Scanner module to NullAwayAnnotatorScanner
**Describe the task**
As @lazaroclapp suggested, the name `Scanner` is too generic, specifically when mentioned along other `ErrorProneCheckers` and should be refined. Currently the suggestion is: `NullAwayAnnotatorScanner`.

**Expected behavior**
Rename of module `Scanner` to `NullAwayAnnotatorScanner` with follow ups of all required updates.

**Screenshots**
Not-Applicable.

**Additional context**
Not-Applicable.

nimakarimipour:
This PR resolves #25 and renames `Scanner` module to `AnnotatorScanner` module.

The name `Scanner` is too generic, specifically when mentioned along other `ErrorProneCheckers` and should be refined. Currently the suggestion is: `AnnotatorScanner`.

This involved changes in the configurations, unit tests project templates, and install scripts.

@lazaroclapp Previously you suggested `NullAwayAnnotatorScanner`, however, this checker is not related to NullAway and I am intending to apply the same technique and use this whole repo in general to infer other type systems. Given that this repo is called `NullAwayAnnotator`, I still think it makes sense to generalize this name. Please let me know your thoughts on this and if you think I should rename it to `NullAwayAnnotatorScanner`. Thanks :)

----
**Update**:
The final decision regarding name is: `TypeAnnotatorScanner`.


lazaroclapp:
Haven't looked at the specific rename changes, but regarding the `AnnotatorScanner` name itself...

> Previously you suggested `NullAwayAnnotatorScanner`, however, this checker is not related to NullAway and I am intending to apply the same technique and use this whole repo in general to infer other type systems. Given that this repo is called `NullAwayAnnotator`, I still think it makes sense to generalize this name.

My problem here is not really within the context of this repo. `Scanner` would be fine if the checker name only ever appeared in the context of this repository or as dependency artifact name (i.e. `edu.ucr.cs.riple.nullawayannotator:scanner:$someVersion`). The problem is that the simple name is also used as the name of the checker in Error Prone: `-Xep:AnnotatorScanner:[...]` becomes part of our configuration, which is file configuring a few dozen checkers - most of them part of Error Prone itself. So my concern is being able to distinguish it from some internal EP checker within that list or (more generally and less Uber specific!) within the arguments passed to `javac`. Clearly this isn't part of some general Error Prone Annotator, at least not for the time being.

I would consider `UCRAnnotatorScanner`, if you don't want to tie it too closely to NullAway. That said, my hope is still that we can get to the point of merging all pieces of the auto-annotator into the uber/NullAway repo at some point and release them together, so, not sure you really can or want to distance yourself from the NullAway brand here. Thoughts, @msridhar ?

Another thing we could do is ask Error Prone folks (i.e. raise an issue on their tracker) about whether there is any sort of name-spacing convention that is used or can be used for third-party Error Prone checkers. But I, personally, am not aware of one.

runner/README.md
```diff
@@ -11,15 +11,15 @@ Below are the instructions to prepare the target project:
...
+2. `Annotator Scanner` checker must be activated with version >= `1.2.6-LOCAL`, see more about `Scanner` [here](../scanner/README.md).
 
 #### Error Prone Flags
 ```
 "-Xep:NullAway:ERROR", // to activate NullAway
 "-XepOpt:NullAway:SerializeFixMetadata=true",
 "-XepOpt:NullAway:FixSerializationConfigPath=path_to_nullaway_config.xml",
-"-Xep:Scanner:ERROR", // to activate Scanner
-"-XepOpt:Scanner:ConfigPath=path_to_scanner_config.xml",
+"-Xep:AnnotatorScanner:ERROR", // to activate Annotator AnnotatorScanner
```
ketkarameya, line 21:
```suggestion
"-Xep:AnnotatorScanner:ERROR", // to activate AnnotatorScanner
```

README.md
```diff
@@ -113,14 +113,14 @@ public class Test {
...
-* [Scanner](scanner/README.md) module should be installed in the ```local maven``` repository.
+* [Annotator Scanner](annotator-scanner/README.md) module should be installed in the ```local maven``` repository.
     * To install it, run `./gradlew install`
 
 * Jar file of [Core](core/README.md) module should be located at `runner/jars`.
     * To create the jar file run: `./gradlew fatjar`
 
 A script is provided which executes all installation tasks above. 
-It installs `Scanner` checker in `maven local` and creates the jar file of `core` module and moves it to the correct path. To run it, execute the command below:
+It installs `Annotator Scanner` checker in `maven local` and creates the jar file of `core` module and moves it to the correct path. To run it, execute the command below:
```
ketkarameya, line 123:
```suggestion
It installs `AnnotatorScanner` checker in `maven local` and creates the jar file of `core` module and moves it to the correct path. To run it, execute the command below:
```

ketkarameya (approved):
LGTM

msridhar:
Hey, so bottom line, I think we need a better name than `AnnotatorScanner`.  What about `CodeIndexSerializer`?  I think what this checker does is serialize some indexing information about the code, like where fields are used, where methods are called, etc.  Or maybe `TypeAnnotatorIndexer`?  @lazaroclapp we do have concrete plans to apply this technique to more than just NullAway, so I think keeping the naming separate and possibly the repositories may be a good idea for now, at least until we can have a more detailed discussion about future plans.

msridhar (dismissed):
Would like the naming discussion resolved before this lands, if possible

runner/README.md
nimakarimipour: 54540be0a34cd643e6b8cf733250e9a8de7fbfd7

README.md
nimakarimipour: 54540be0a34cd643e6b8cf733250e9a8de7fbfd7

msridhar (commented):
No time to review and approve all these changes but I don't object

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
```diff
@@ -118,6 +118,7 @@ public void testRequiredFlagsMissingCli() {
               IllegalArgumentException.class,
               () -> new Config(makeCommandLineArguments(incompleteFlags)));
       // Check the error message.
+      System.out.println(ex.getMessage());
```
lazaroclapp, line 121: Was this for debugging? I don't think the tests should print to stdout on every run, and we are already checking the message below on a successful run. Let's remove this line.

library-model-loader/src/main/resources/edu/ucr/cs/riple/librarymodel/nullable-methods.tsv
```diff
@@ -1,2 +0,0 @@
-test.target.Foo	returnNullableBad(int)
-test.target.Foo	returnNullableGood(int)
```
lazaroclapp, line 2: Why are re removing this library model TSV? (Just want to understand, since it seems unrelated to the renaming and it isn't called out in the PR description)

runner/start.sh
```diff
@@ -1,12 +1,26 @@
...
+  popd
+popd
+
+pushd "$PROJECT_ROOT"
+  pushd runner
+    pushd "$DIR"
+      java -jar core.jar "$@"
+    popd
+  popd
+popd
```
lazaroclapp, line 26: Why isn't this ending in a newline?

type-annotator-scanner/README.md
```diff
@@ -0,0 +1 @@
+## Type Annotator Scanner (Scanner)
```
lazaroclapp, line 1: At some point we might want proper documentation here. Even now, I wouldn't mind a single-paragraph description of what the Type Annotator Scanner does and what it is used for here 🙂 (don't spend a lot of time on it, but it's certainly worth 5-10 minutes to explain its overall purpose)

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
nimakarimipour: Yes exactly, this is left out code from debugging. bd21c3d0e4abd6d8e120b7bc20523056e5187e0b

library-model-loader/src/main/resources/edu/ucr/cs/riple/librarymodel/nullable-methods.tsv
nimakarimipour: `LibraryModelsLoader` module here is used for unit tests and these are left over contents from previous runs. They shouldn't appear in the git repository.

runner/start.sh
nimakarimipour: 8f79e985a4d0f416d63cd465a247ac3ffe11a573

type-annotator-scanner/README.md
nimakarimipour: Sure, I added an overview of the module. Will complete it later :)
048a3bb4e86ee5417677840774efb91dda4f5c70

type-annotator-scanner/README.md
```diff
@@ -1 +1,100 @@
...
+4. Annotation of the method return.
+5. Unique id of this node.
+6. Unique id of parent node.
+7. Visibility of `m`.
+8. Return type of `m`.
+
+#### Class info
+Serializes flat name of all observed classes and uri of the file containing the class.
+
+These data is used by the `core` module to detect multiple inline field declarations.
```
lazaroclapp, line 41: "This data" (uncountable noun)

lazaroclapp (approved):
LGTM! (After the change to the TSV file lands, can we make a follow up PR to `.gitignore` that file? Or otherwise make sure that the tests clean up their changes to that file after running, whether or not they succeed?)

nimakarimipour:
> Other than a minor doc nit below, this LGTM! (After the change to the TSV file lands, can we make a follow up PR to `.gitignore` that file? Or otherwise make sure that the tests clean up their changes to that file after running, whether or not they succeed?)

Thank you for the review. Sure will resolve the `tsv` file in a follow up PR.

type-annotator-scanner/README.md
nimakarimipour: 1e2286bae6c6cc4f36c86d81d760df0f80c294fd

nimakarimipour:
@ketkarameya I can see you already approved this PR. Will land it as soon as the checks pass. Thank you.

PR #39, opened 2022-08-09
Fix #38, Add UUID to every config file

### LINKED ISSUE #38: Build requests from Annotator may not execute due to build systems caching strategy.
**Describe the bug**
In environments where the build system is extensively using caching strategies, a build request from Annotator may not execute if no file is modified under the module directory. Even if config files of annotator are written inside the module, since config files content can be identical within build request, it might hit the cache and the using build system will not rebuild the module. If the project is not rebuilt, annotator will not correctly compute the effectiveness of fixes and the serialization outputs may not be even located.

**To Reproduce**
Re-use the AutoAnnotator

**Expected behavior**
When ever annotator is requesting a build, all `Error Prone` checkers should re-execute.

**Screenshots**
Not-Applicable.

 - OS: Linux
 - Annotator Version: v1.3.2-SNAPSHOT

**Additional context**
Not-Applicable.

nimakarimipour:
This PR resolves #38 by adding a `UUID` to every generated config file to guarantee with same configuration values, content of the file is still modified. This will trigger build systems to rebuild the module whenever annotator requests rebuilds.

Please note: `UUID` alone might not be effective unless the build systems takes the config files states into account in their caching strategies.

### DIFF
--- a/core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
@@ -47,6 +47,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.xml.parsers.DocumentBuilder;
@@ -179,6 +180,11 @@ public class Utility {
       outputDir.setTextContent(config.outputDirectory);
       rootElement.appendChild(outputDir);
 
+      // UUID
+      Element uuid = doc.createElement("uuid");
+      uuid.setTextContent(UUID.randomUUID().toString());
+      rootElement.appendChild(uuid);
+
       // Writings
       TransformerFactory transformerFactory = TransformerFactory.newInstance();
       Transformer transformer = transformerFactory.newTransformer();
@@ -229,6 +235,11 @@ public class Utility {
       Element outputDir = doc.createElement("path");
       outputDir.setTextContent(info.dir.toString());
       rootElement.appendChild(outputDir);
+
+      // UUID
+      Element uuid = doc.createElement("uuid");
+      uuid.setTextContent(UUID.randomUUID().toString());
+      rootElement.appendChild(uuid);
 
       // Writings
       TransformerFactory transformerFactory = TransformerFactory.newInstance();
--- a/core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
+++ b/core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
@@ -107,7 +107,7 @@ public class ConfigurationTest {
   }
 
   @Test
-  public void test_required_flags_missing_cli() {
+  public void testRequiredFlagsMissingCli() {
     // Check if each is missed.
     for (int i = 0; i < requiredFlagsCli.size(); i++) {
       List<CLIFlag> incompleteFlags = new ArrayList<>(requiredFlagsCli);
@@ -123,7 +123,7 @@ public class ConfigurationTest {
   }
 
   @Test
-  public void test_required_flags_cli() {
+  public void testRequiredFlagsCli() {
     Config config = new Config(makeCommandLineArguments(requiredFlagsCli));
     assertEquals("./gradlew compileJava", config.buildCommand);
     assertEquals(testDir, config.globalDir);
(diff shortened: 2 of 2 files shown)


lazaroclapp (approved):
Makes sense to me! It's a bit of a hack and might result in redundant builds that could be avoided if we make our analysis outputs part of the build outputs for every build system we integrate with (and thus part of what gets returned by the build cache), but that's a complex build-system dependent optimization and, in theory, given how the auto-annotator coordinates builds, such redundant builds should be relatively rare, correct?

nimakarimipour:
> Makes sense to me! It's a bit of a hack and might result in redundant builds that could be avoided if we make our analysis outputs part of the build outputs for every build system we integrate with (and thus part of what gets returned by the build cache), but that's a complex build-system dependent optimization and, in theory, given how the auto-annotator coordinates builds, such redundant builds should be relatively rare, correct?

Sure, the approach you mentioned sounds really interesting will definitely work on that in future. Yes they are relatively rare, but using `UUID` made it very stable so far.

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
```diff
@@ -160,6 +175,61 @@ public void test_required_flags_for_downstream_dependency_analysis_cli() {
     assertEquals(downstreamInfo, config.downstreamInfo);
   }
 
+  @Test
+  public void test_config_files_have_different_uuid() {
```
ketkarameya, line 179:
```suggestion
  public void testConfigFilesHaveDifferentUUID() {
```

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
```diff
@@ -160,6 +175,61 @@ public void test_required_flags_for_downstream_dependency_analysis_cli() {
...
+      DocumentBuilder builder = factory.newDocumentBuilder();
+      Document doc = builder.parse(Files.newInputStream(path));
+      doc.normalize();
+      XPath xPath = XPathFactory.newInstance().newXPath();
+      return ((Node) xPath.compile(key).evaluate(doc, XPathConstants.NODE)).getTextContent();
+    } catch (XPathExpressionException
+        | ParserConfigurationException
+        | IOException
+        | SAXException ex) {
+      throw new RuntimeException("Could not extract value from tag: " + key, ex);
```
ketkarameya, line 229: does adding a comma between key and ex imply they will be concatenated with a space between them?

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
nimakarimipour: d8227e8ba51b09112b2956aac622e5f498637618

Thanks for the suggestion, I will rename all other test methods outside of scope of this PR in a separate PR.

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
nimakarimipour: I am not sure if I fully understood your comment, `ex` will be here will be passed as the cause of the exception and is not part of the string.

ketkarameya: Oh lol. Forget it.

ketkarameya (approved):
lgtm.

core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
```diff
@@ -179,6 +180,11 @@ public static void writeNullAwayConfigInXMLFormat(FixSerializationConfig config,
       outputDir.setTextContent(config.outputDirectory);
       rootElement.appendChild(outputDir);
 
+      // UUID
```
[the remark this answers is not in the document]
nimakarimipour: I wanted to do that but I think it might make it complicated, please consider the code below:
```java
Element suggestElement = doc.createElement("suggest");
suggestElement.setAttribute("active", String.valueOf(config.suggestEnabled));
suggestElement.setAttribute("enclosing", String.valueOf(config.suggestEnclosing));
rootElement.appendChild(suggestElement);
```

If we follow the extract method, we should have a method with signature `addElementWithAttribute(Document, ElementName, TagName, TagValue)` which with the above code is not possible to use this method since it is adding multiple tags to the same element so it either has to look for an existing element with each call (and only add it to `rootElement` as a child if its not added before) or pass a `Map` of tag name/value to the `addElementWithAttribute` method to support the code above.  Which with both version, I think reading the code at the current state will give an idea how the `xml` will look like clearer.

nimakarimipour:
> lgtm.

Thank you very much, will land it now :)

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
lazaroclapp: No. This is attaching `ex` as the cause of the `RuntimeException`, see the docs [here](https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html#RuntimeException(java.lang.String,%20java.lang.Throwable)). This is a common pattern in Java and means that both the stack trace starting at this line, and that of the original exception (plus its error message) will be preserved.

PR #41, opened 2022-08-12
Add Error Prone checks and ban declaring mutable static fields.

### LINKED ISSUE #40: Crash: ConcurrentModificationError
**Describe the bug**
In environment where multiple instances of `Scanner` is running, it is possible that the tool crashes in error prone checks because of mutable static fields. The exception is `ConcurrentModificationError`

**To Reproduce**
Run multiple instance of `Scanner` in parallel.

**Expected behavior**
Stable run of multiple `Scanner` instances.

**Stack trace**
```java
java.util.ConcurrentModificationException
  	at java.base/java.util.HashMap$KeySpliterator.tryAdvance(HashMap.java:1644)
  	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127)
  	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502)
  	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
  	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
  	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
  	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
  	at java.base/java.util.stream.ReferencePipeline.findAny(ReferencePipeline.java:548)
  	at edu.ucr.cs.riple.scanner.out.MethodInfo.findOrCreate(MethodInfo.java:65)
  	at edu.ucr.cs.riple.scanner.out.MethodInfo.findParent(MethodInfo.java:89)
  	at edu.ucr.cs.riple.scanner.Scanner.matchMethod(Scanner.java:106)
  	at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:451)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitMethod(ErrorProneScanner.java:741)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitMethod(ErrorProneScanner.java:152)
        ......
```

 - OS: OS X
 - Version 12.5
 - Annotator Version: v1.3.2-SNAPSHOT

**Additional context**
Not-Applicable.

nimakarimipour:
This PR resolves #40 by adding a local `Error Prone` checker named `BanMutableStatic` that bans declaring any mutable static fields (cover only shallow mutability) in all modules in this repository. 

Along that, it also had general improvement of addressing error prone default checkers reported warnings, mostly related to the categories below:
1. Missing braces (generated codes by Intellij)
2. Missing `@Override`
3. Missing Charset when reading/writing files.
4. Format in javadoc


checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
```diff
@@ -0,0 +1,68 @@
...
+  public final String ERROR_MESSAGE =
+      "Variable %s is declared as a mutable static field, should be replaced with: final static %s %s";
+  private final String LINK_URL = "https://github.com/nimakarimipour/NullAwayAnnotator";
+
+  public BanMutableStatic() {}
+
+  @Override
+  public Description matchVariable(VariableTree tree, VisitorState state) {
+    boolean isStatic = tree.getModifiers().getFlags().contains(Modifier.STATIC);
+    boolean isMutable = !tree.getModifiers().getFlags().contains(Modifier.FINAL);
```
msridhar, line 58: This only covers "shallow" mutability, but "deep" mutability is also important.  E.g., if you have a `static final HashMap` that is mutable even though it is final, since you can mutate the contents of the `HashMap`.  So this check will catch some of the bad cases, but not all.

nimakarimipour: Thanks for the note. Is this something we would like to avoid ? Is this also problematic If multiple instances are running and is guaranteed that the list is singleton and shared among them ?

msridhar: Absolutely we need to avoid it.  It causes huge problems like data races and memory leaks.  I don't understand the singleton list question; if you know ahead of time it's a singleton why do you need a list?

nimakarimipour: Sure, will also cover them. Thank you. By singleton I meant only __one__ instance of list which can have arbitrarily number of elements. I will cover cases you mentioned as well.

msridhar: If the contents of the list change over time then it is bad 🙂 If it's an `ImmutableList` it's ok

lazaroclapp: Haven't had a chance yet to do a full review, but, regarding the `BanMutableStatic` checker only... why not make it so it requires that static fields are both `final` _and_ annotated `@Immutable`, then leverage this existing EP checker: https://errorprone.info/bugpattern/Immutable

msridhar: @lazaroclapp I think the`@Immutable` annotation applies to classes, not fields?  If we really wanted to catch this we could check that the type of the field is `@Immutable` maybe by re-using code from https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/threadsafety/ImmutableAnalysis.java

nimakarimipour: @lazaroclapp @msridhar I looked at this code, all I can access is `checkForImmutability` method which can be called on class tree. Would you please tell me what you mean by reusing this code. I think we have two options, either override match class method and call it from there, or copy codes from `areFieldsImmutable` and keep overriding `matchVariable` method. Thank you.

msridhar: At this point I think this is low priority, and I wouldn't spend time improving the mutable static checker.  Given the size of the scanner, it shouldn't be hard to just search for all static fields using IntelliJ and check them manually.

nimakarimipour: @msridhar Ok sure sounds good. I will keep the checker as is. I may work on it in future if there was nothing outstanding to do since it seems like an interesting side project :).

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Nima Karimipour
```
lazaroclapp, line 4: 2022

checks/ban-mutable-static/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+        classpath 'com.android.tools.build:gradle:4.1.1'
+    }
+}
+
+plugins {
+    id "net.ltgt.errorprone" version "2.0.1" apply false
+    id 'com.github.sherter.google-java-format' version '0.9'
+}
+
+apply plugin: 'com.vanniktech.maven.publish'
```
lazaroclapp, line 41: Do we want to make this publishable in its current form? I am pretty sure the rule only applies to specific codebases, such as other EP checkers. If we wanted this to be generally useful, it would need some sort of annotation check to enable the checker (e.g. a package level `@DisallowMutableStatic` annotation or similar).

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public final String ERROR_MESSAGE =
+      "Variable %s is declared as a mutable static field, should be replaced with: final static %s %s";
+  private final String LINK_URL = "https://github.com/nimakarimipour/NullAwayAnnotator";
+
+  public BanMutableStatic() {}
+
+  @Override
+  public Description matchVariable(VariableTree tree, VisitorState state) {
+    // At this moment, this checker This only covers "shallow" mutability. Will make it cover deep
```
lazaroclapp, line 57: That upper case "This" seems to be a residue from an incomplete comment rewrite.

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
```diff
@@ -0,0 +1,70 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+    name = "BanMutableStatic",
+    altNames = {"BansMutableStaticFields"},
+    summary = "Bans mutable static field declaration",
+    tags = BugPattern.StandardTags.STYLE,
+    severity = SUGGESTION)
+@SuppressWarnings("BugPatternNaming")
+public class BanMutableStatic extends BugChecker implements BugChecker.VariableTreeMatcher {
```
lazaroclapp, line 47: Public class, so it needs a javadoc. That would be a good place to mention the limitations (e.g. only detecting shallow mutability)

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
```diff
@@ -0,0 +1,70 @@
...
+    // At this moment, this checker This only covers "shallow" mutability. Will make it cover deep
+    // cases in future which is not the priority for now.
+    boolean isStatic = tree.getModifiers().getFlags().contains(Modifier.STATIC);
+    boolean isMutable = !tree.getModifiers().getFlags().contains(Modifier.FINAL);
+    if (isMutable && isStatic) {
+      Symbol symbol = ASTHelpers.getSymbol(tree);
+      Description.Builder builder = buildDescription(tree);
+      builder.setMessage(String.format(ERROR_MESSAGE, symbol, symbol.type, symbol));
+      builder.setLinkUrl(LINK_URL);
+      return builder.build();
```
lazaroclapp, line 66: Nit: This is fine, but it would be more idiomatic when using a builder to do:

```
return buildDescription(tree)
     .setMessage(String.format(ERROR_MESSAGE, symbol, symbol.type, symbol))
     .setLinkUrl(LINK_URL)
     .build();
```

checks/ban-mutable-static/src/test/java/edu/ucr/cs/riple/banmutablestatic/CoreTest.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2020 Nima Karimipour
```
lazaroclapp, line 4: Wrong year.

core/src/main/java/Main.java
```diff
@@ -25,6 +25,7 @@
 import edu.ucr.cs.riple.core.Annotator;
 import edu.ucr.cs.riple.core.Config;
 
+@SuppressWarnings("DefaultPackage")
```
lazaroclapp, line 28: Why suppressing instead of making this `package edu.ucr.cs.riple.core;` and changing whatever command is used to run it or the jars metadata?

core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -403,10 +410,21 @@ public Config(String configPath) {
...
     log.reset();
   }
 
+  /**
+   * Returns the latest id associated to a module, used to create unique ids for each module and
+   * increments it.
+   *
+   * @return last id value used in integer.
+   */
+  public int getNextModuleUniqueID() {
```
lazaroclapp, line 424: I see no uses of this method outside this class itself. Given that, it should be private, but, since it additionally is a one-liner, why aren't we just using `moduleCounterID++` in the two places we call `getNextModuleUniqueID()` and just removing this method?

core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -559,10 +577,11 @@ public void write(Path path) {
         downstreamDependency.put("BUILD_COMMAND", downstreamBuildCommand);
       }
       json.put("DOWNSTREAM_DEPENDENCY_ANALYSIS", downstreamDependency);
-      try (FileWriter file = new FileWriter(path.toFile())) {
+      try (BufferedWriter file =
+          Files.newBufferedWriter(path.toFile().toPath(), Charset.defaultCharset())) {
         file.write(json.toJSONString());
       } catch (IOException e) {
-        e.printStackTrace();
+        System.err.println("Error happened in writing config json, " + e);
```
lazaroclapp, line 584: Why coma instead of `:`?

core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldInitializationAnalysis.java
```diff
@@ -71,6 +71,7 @@ public Stream<OnMethod> findInitializers(Set<OnField> uninitializedFields) {
   }
 
   /** Stores class field / method initialization status. */
+  @SuppressWarnings("JavaLangClash")
```
lazaroclapp, line 74: Please solve in a follow up diff. One option is to use `Klass` or `ClassRecord`

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -26,6 +26,7 @@
 import java.util.Objects;
 
 /** Represents an error reported by NullAway. */
+@SuppressWarnings("JavaLangClash")
```
lazaroclapp, line 29: As above: no need to solve in this PR, but let's fix this in a follow up.

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Index.java
```diff
@@ -105,7 +105,7 @@ public void index() {
               line = br.readLine();
             }
           } catch (IOException e) {
-            e.printStackTrace();
+            throw new RuntimeException("Error happened in indexing", e);
```
lazaroclapp, line 108: 👍

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnField.java
```diff
@@ -45,7 +45,7 @@ public OnField(String uri, String clazz, Set<String> variables) {
 
   @Override
   public Location duplicate() {
-    return new OnField(clazz, uri, new HashSet<>(variables));
+    return new OnField(uri, clazz, new HashSet<>(variables));
```
lazaroclapp, line 48: Interesting find... was this spotted by an Error Prone check?

library-model-loader/build.gradle
```diff
@@ -35,3 +35,15 @@ dependencies {
...
+
+tasks.withType(JavaCompile) {
+    // remove the if condition if you want to run NullAway on test code
+    if (!name.toLowerCase().contains("test")) {
+        options.errorprone {
+            check("NullAway", CheckSeverity.ERROR)
+            option("NullAway:AnnotatedPackages", "edu")
+        }
+    }
+}
```
lazaroclapp, line 49: Please end file in new line

library-model-loader/build.gradle
```diff
@@ -35,3 +35,15 @@ dependencies {
...
 }
+
+import net.ltgt.gradle.errorprone.CheckSeverity
+
+tasks.withType(JavaCompile) {
+    // remove the if condition if you want to run NullAway on test code
+    if (!name.toLowerCase().contains("test")) {
+        options.errorprone {
+            check("NullAway", CheckSeverity.ERROR)
+            option("NullAway:AnnotatedPackages", "edu")
```
lazaroclapp, line 46: A bit aggressive. This should be at most `edu.ucr`, and arguably `edu.ucr.cs.riple`, I think (cc: @msridhar ), otherwise you are relying on no dependencies ever being `edu.*`

library-model-loader/src/main/resources/edu/ucr/cs/riple/librarymodel/nullable-methods.tsv
```diff
@@ -1,2 +0,0 @@
-test.target.Foo	returnNullableBad(int)
-test.target.Foo	returnNullableGood(int)
```
lazaroclapp, line 2: Why are these being removed as part of this PR again?

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
```diff
@@ -53,6 +56,26 @@ public interface Config {
   @Nonnull
   Path getOutputDirectory();
 
+  Context getContext();
+
+  class Context {
```
lazaroclapp, line 61: I like the general idea here, which is that we are passing our own "VisitorState" of sorts. That said, it feels a bit strange to be passing the "context" inside the "configuration", rather than the other way around (the configuration being part of the context in which we are running). How massive of a refactor would be to have an `ScannerContext` top level object that gets passed around and has `getConfig()` in addition to `getNextMethodId()` (let's not use `getNextUniqueID()` unless we plan to use it across different types of objects)

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
```diff
@@ -53,6 +56,26 @@ public interface Config {
...
+    public final Set<MethodInfo> methodIs;
```
lazaroclapp, line 62: Also, no idea why we are collecting this set of method info records here, as part of `Context`, but either way this isn't a very descriptive field name. Maybe `methodInfoRecords`?

scanner/src/main/java/edu/ucr/cs/riple/scanner/Scanner.java
```diff
@@ -54,6 +54,7 @@
     summary = "Serialized type based call/field graph.",
     tags = BugPattern.StandardTags.STYLE,
     severity = SUGGESTION)
+@SuppressWarnings("BugPatternNaming")
```
lazaroclapp, line 57: Is the warning here that we don't need `name = [...]` if its the same name as the class? If so, why suppress instead of removing `name`? Or is the error something else?

library-model-loader/build.gradle
lazaroclapp: Edit: Note that what I am suggesting here would require all test examples to be inside a package like `edu.ucr.cs.riple[.scanner.test]`, but that is probably best practice anyways.

settings.gradle
```diff
@@ -27,4 +27,6 @@ include 'core'
 include 'scanner'
 include 'injector'
 include 'library-model-loader'
+include 'checks'
```
lazaroclapp, line 30: What's under `:checks`? The only `build.gradle` I see is under `checks/ban-mutable-static`.

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
nimakarimipour: 795ee5d04617969ef9582879b47d65ae1e0f18bd

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
nimakarimipour: 3ab8024c1ca2363636c694ee0df8e091ae013dc5

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
nimakarimipour: 462343002666529bc44664a74f22b1cc79a51a24

checks/ban-mutable-static/src/main/java/edu/ucr/cs/riple/banmutablestatic/BanMutableStatic.java
nimakarimipour: c6bae1e28b8cd37316f4e5ed1a147eb4e76b668f

checks/ban-mutable-static/src/test/java/edu/ucr/cs/riple/banmutablestatic/CoreTest.java
nimakarimipour: ccb2db8ddef57ef80bcda78253bd80a075a68566

core/src/main/java/edu/ucr/cs/riple/core/Config.java
nimakarimipour: This method is called in the `scanner.out.MethodInfo.java` in the first line of the constructor. Do you still think I should remove this method ?

core/src/main/java/edu/ucr/cs/riple/core/Config.java
nimakarimipour: That was a mistake, thanks for noticing this.
e33fde946908b14e6037a5084657397176de1c36

core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldInitializationAnalysis.java
nimakarimipour: Sure, `ClassRecord` sounds really good to me.

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnField.java
nimakarimipour: Yes :) but this method is not used anymore.

library-model-loader/build.gradle
nimakarimipour: 5802e7e7ec2c5d251813668e0a58f86204ddd1d9

library-model-loader/src/main/resources/edu/ucr/cs/riple/librarymodel/nullable-methods.tsv
nimakarimipour: `LibraryModelsLoader` module here is used for unit tests and these are left over contents from previous runs. They shouldn't appear in the git repository.

lazaroclapp: Ok, does that mean we are missing a `.gitignore` entry somewhere?

nimakarimipour: I thought maybe we can keep the file which is more clearer for the reader of the code where library models loader read the data from. `gitignore` will delete the file I guess. There is another approach that skips the file from indexing but that will happen only in my repository not for all clones.

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
nimakarimipour: This is used to not to increment unique id for an already visited method. It is currently `visitedMethods` please let me know if you still think `methodInfoRecords` is more descriptive.

scanner/src/main/java/edu/ucr/cs/riple/scanner/Scanner.java
nimakarimipour: Yes that error is due to the issue you mentioned, but there is no constructor which does not require the name. When I did that, there was an error that name is missing.

library-model-loader/src/main/resources/edu/ucr/cs/riple/librarymodel/nullable-methods.tsv
lazaroclapp: If the file is already landed, I don't believe `.gitignore` will remove it, just prevent changes to it from being added to new commits. That said, the alternative is to make sure that, when tests run, whether or not they succeed, the end state of this file is empty.

library-model-loader/build.gradle
lazaroclapp: @nimakarimipour No action here? Why?

scanner/src/main/java/edu/ucr/cs/riple/scanner/ScannerContext.java
```diff
@@ -0,0 +1,96 @@
...
+import com.google.common.collect.MultimapBuilder;
+import edu.ucr.cs.riple.scanner.out.MethodInfo;
+import java.util.stream.Stream;
+
+/**
+ * Container class to store the state of the checker.
+ *
+ * <p>Note: We could not rely on the given {@link com.google.errorprone.VisitorState#context} since
+ * we receive different instances of {@link com.google.errorprone.SubContext} and writes on these
+ * instances are not globally reflected.
```
lazaroclapp, line 37: 👍

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
```diff
@@ -157,4 +159,15 @@ private String getVisibilityOfMethod() {
     }
     return "package";
   }
+
+  /**
+   * Calculates hash. This method is used outside this class to calculate the expected hash based on
+   * instance's properties value if the actual instance is not available.
```
lazaroclapp, line 165: This doc could be made a bit clearer, perhaps by pointing out that the hash is calculated based on a `MethodSymbol` instance since we want to predict the hash of a potential `MethodInfo` object without creating the `MethodInfo` instance?

checks/ban-mutable-static/build.gradle
lazaroclapp: @nimakarimipour ? Thoughts here?

core/src/main/java/Main.java
lazaroclapp: @nimakarimipour ?

core/src/main/java/edu/ucr/cs/riple/core/Config.java
lazaroclapp: How? This is a method of the `Config` class within `core`... the method called inside `MethodInfo` in scanner is `ScannerContext::getNextUniqueID()`

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnField.java
lazaroclapp: Then why do we keep it, rather than cleaning it up altogether? Or is it just not used in the default configuration but still something we want to support?

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
lazaroclapp: My suggestion to rename it to `getNextMethodId()` depends on whether the ID is used to number things other than methods.

The method I think should be removed is not this, but `edu.ucr.cs.riple.core.Config.getNextModuleUniqueID()` (As a test, I am pretty sure if you make that one private, nothing is calling it outside of `edu.ucr.cs.riple.core.Config` :/

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
lazaroclapp: I am fine with `visitedMethods`

scanner/src/main/java/edu/ucr/cs/riple/scanner/Scanner.java
lazaroclapp: Weird, I wonder if it's a discrepancy between `@BugPattern` requirements across different EP versions... is it the same version of error prone that failes with and without the name locally or are they different CI tasks?

lazaroclapp (commented):
Did another pass, Nima. but I think you missed a few comments from the previous pass (or at least I don't see any change in response to them, nor any reply. I just responded to them to bump them)

scanner/src/main/java/edu/ucr/cs/riple/scanner/Scanner.java
msridhar: We have this suppression in NullAway as well:

https://github.com/uber/NullAway/blob/master/nullaway/src/main/java/com/uber/nullaway/NullAway.java#L150

There it was because we want to be able to build against Error Prone versions older than 2.11, and there was some issue with setting the bug pattern name consistently in older versions (I can't remember all the details).

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
nimakarimipour: @lazaroclapp Thank you for the clarification, I mistakenly thought you are talking about `getNextUniqueID` in `Scanner` config. Changed the visibility from `public` to `private`. 26083ada5e91d7ef4b8cc3ab3bbe994c51c68599

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnField.java
nimakarimipour: Actually this method is still called in the tests, there was a usage for this in previous versions of `injector` where we had a flag to distinguish `add` from `remove` annotation request. But at this moment, we have a clear types of `RemoveAnnotation` and `AddAnnotation` that extends the `Change` class. With this design we no longer use `duplicate()` method in `core` module to remove the injected annotations. However, this method is still used in a test. (The test checks we do not inject multiple `@Nullable` annotations on the same element once one is already injected)

core/src/main/java/Main.java
nimakarimipour: Moved it to the `package edu.ucr.cs.riple.core`. d023ea6635a4652f59a420cf1fb8f449209099cc

checks/ban-mutable-static/build.gradle
nimakarimipour: I actually removed the publication feature from this checker as it is only useful internally for now and it found all mutable static fields (shallow cases). However, I am interested into completing this in future to cover deep cases as I find it a very useful checker honestly. Maybe we can complete this later in a follow up PRs and cover deep cases as well.

c85d32e3adbf73b23121b4a25d819b78ea4dda5b

library-model-loader/build.gradle
nimakarimipour: Sorry for missing this comment and thank you for noticing this. I actually wrote "edu" just because there wasn't significant code in this library and thought it only affects the code available in `src` directory which is not true.

I also inspired from `NullAway Library Models` sample. Now that I have a much better understanding of how these works and the correct configurations, I see that this was only there as the whole mission of enrolling NullAway into NullAway and library models loader does not need `NullAway` to be on it's annotation processor. Removed it. 01a15119f5e34bbbc475210b92e9ed390f501ba6

library-model-loader/src/main/resources/edu/ucr/cs/riple/librarymodel/nullable-methods.tsv
nimakarimipour: Sure will do that.

core/src/main/java/edu/ucr/cs/riple/core/Config.java
nimakarimipour: Thank you for the clarification, I mistakenly thought you are talking about `getNextUniqueID` in Scanner config. Changed the visibility from public to private. https://github.com/nimakarimipour/NullAwayAnnotator/commit/26083ada5e91d7ef4b8cc3ab3bbe994c51c68599

scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
nimakarimipour: 806f60d9bdb8d2793fe9525312a3c0e4f78ef22e

scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
nimakarimipour: 0b025195ebd9f4a8c53c69ad3ea34db3e77cf181

library-model-loader/build.gradle
```diff
@@ -32,18 +32,5 @@ dependencies {
...
-
-tasks.withType(JavaCompile) {
-    // remove the if condition if you want to run NullAway on test code
-    if (!name.toLowerCase().contains("test")) {
-        options.errorprone {
-            check("NullAway", CheckSeverity.ERROR)
-            option("NullAway:AnnotatedPackages", "edu")
-        }
-    }
-}
```
lazaroclapp: Are we entirely removing running NullAway on this target here? Or do we get this from some other more general task in the build? (I think we shouldn't be running for `edu.*` in general here, but we should still be running for `edu.ucr.cs.riple.*`, no?)

lazaroclapp (commented):
One quick note below, and also I wonder if you have any thoughts on the comment around `OnField.duplicate(...)`?

library-model-loader/build.gradle
nimakarimipour: Hi @lazaroclapp I replied to your comment around `OnField.duplicate()` previously. Please find it [here](https://github.com/nimakarimipour/NullAwayAnnotator/pull/41#discussion_r947204318).
Regarding the `NullAway` enrollment, yes we need to run it under package name `edu.ucr.cs.riple.*` and I think we should plan to do this for the whole repo at some point. That enrollment, which only covers `LibraryModelsLoader` was due to my assumption that in order for `LibraryModelsLoader` to be picked up by `NullAway`, it should have `NullAway` in its annotation processor path which is not true and It wasn’t part of any `NullAway` enrollment plan. 
That's why I removed it for now. Will enroll all modules in NullAway under the name `edu.ucr.cs.riple.*` in near future.

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnField.java
lazaroclapp: And yet nothing was broken with that test due to the parameters being reversed here? Um... either way, agree at this point that this is separate from this PR at any rate :)

library-model-loader/build.gradle
lazaroclapp: Ok, this all makes sense! Sorry about the confusion with `OnField.duplicate()`, GitHub seems to somehow split the thread when it includes comments that are part of two reviews 😓

lazaroclapp (approved):
Given the latest clarifications, this change LGTM! 🎉

core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
```diff
@@ -307,14 +306,12 @@ public static ProgressBar createProgressBar(String taskName, int steps) {
   }
 
   public static void writeLog(Config config) {
-    File file = config.globalDir.resolve("log.txt").toFile();
-    try (FileOutputStream fos = new FileOutputStream(file)) {
-      BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(fos));
+    Path path = config.globalDir.resolve("log.txt");
+    try (BufferedWriter bw = Files.newBufferedWriter(path, Charset.defaultCharset()); ) {
```
ketkarameya, line 310: Unless we are doing it for some performance reasons, you could use the `Files` api too. `Files.write(...)`

core/src/test/java/edu/ucr/cs/riple/core/tools/CoreTestHelper.java
```diff
@@ -198,16 +198,17 @@ private void compare(Collection<Report> actualOutput) {
 
   private void makeAnnotatorConfigFile(Path path) {
     Config.Builder builder = new Config.Builder();
+    final int[] id = {0};
     builder.configPaths =
         modules.stream()
             .map(
                 name ->
                     new ModuleInfo(
+                        id[0]++,
```
ketkarameya, line 207: side effect inside a lambda :|

core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
nimakarimipour: No it is not for performance, rewrote it with `Files` api. 48d4a6470593935cbb2025c6fd13848f21344672

ketkarameya (approved):

ketkarameya (approved):
There are some nits I left. Overall I think , Lazaro knows the business logic in detail. And I see he has approved.  LGTM

nimakarimipour:
> There are some nits I left. Overall I think , Lazaro knows the business logic in detail. And I see he has approved. LGTM

Thank you for the review, will land it as soon as the checks passes.

core/src/test/java/edu/ucr/cs/riple/core/tools/CoreTestHelper.java
lazaroclapp: Streams and lambdas in Java are not meant to be pure, though. There is a reason the `.forEach(...)` operator exists, which would be pointless in a functional language. Obviously it is a 'use this with care' situation, but incrementing a counter in a sequential stream is not an unreasonable thing to do (or even in a parallel stream, though that would need an atomic counter 🙂 ).

PR #42, opened 2022-08-14
Remove downstream dependency analysis from loop

nimakarimipour:
This PR removes the main part of the downstream dependency analysis from the iteration loop since the state of the downstream dependency module is not changing along the iterations. All computations are valid along the iteration can be reused.

**Note**: This assumption is valid as long as we are analyzing the whole set of public methods on downstream dependencies, by the time this changes (e.g. only analyzing methods that `NullAway` is suggesting to be `@Nullable`) we have to perform the analysis in each iteration.

### DIFF
--- a/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -98,7 +98,11 @@ public class Annotator {
     FieldDeclarationAnalysis fieldDeclarationAnalysis = new FieldDeclarationAnalysis(config.target);
     MethodInheritanceTree tree =
         new MethodInheritanceTree(config.target.dir.resolve(Serializer.METHOD_INFO_FILE_NAME));
-    // It analyzes effects of all public APIs, does not need to re-compute them per iteration.
+    // downStreamDependencyExplorer analyzes effects of all public APIs on downstream dependencies.
+    // Through iterations, since the source code for downstream dependencies does not change and the
+    // computation does not depend on the changes in the target module, it will compute the same
+    // result in each iteration, therefore we perform the analysis only once and reuse it in each
+    // iteration.
     DownStreamDependencyExplorer downStreamDependencyExplorer =
         new DownStreamDependencyExplorer(config, tree);
     if (config.downStreamDependenciesAnalysisActivated) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -96,6 +96,14 @@ private void explore() {
     Utility.buildTarget(config);
     Utility.setScannerCheckerActivation(config.target, false);
     FieldDeclarationAnalysis fieldDeclarationAnalysis = new FieldDeclarationAnalysis(config.target);
+    MethodInheritanceTree tree =
```
ketkarameya, line 99: Do we have any test for this `explore` method? Like i think it has like important logic that needs to have some tests around it.

lazaroclapp: Presumably it gets exercised as part of the unit/integration [tests](https://github.com/nimakarimipour/NullAwayAnnotator/blob/master/core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java#L32) for the full downstream dependency analysis? Or are you thinking of a more targeted way to test `explore()` in isolation? (Presumably mocking the actual compilation task results?)

ketkarameya: I was wondering if there are any E2E test cases that are affected due to this change.  Does the change to the `.tsv` file update the E2E test case? 

Also in general, I saw that this `explorer` method is calling a bunch of APIs -> `ExaustiveExplorer`, `OptimisedExplorer`,`Bank` .. and it is doing a lot of things. Mocking and testing all of this would be too much effort (unless there is a way of smartly extracting out the core exploration functionality and testing it in isolation), 
. At least for now, It would be either helpful have a a method level comment describing the functionality of this method or line-by-line comments .

nimakarimipour: @lazaroclapp @ketkarameya Sure I agree that it needs a unit test to check functionality of `explorers` in a finer granularity level and you are right it needs a significant effort to mock this. The main reason for that is `explorers` require a command to re-run the analysis (NullAway at this moment) and mocking that requires a lot of work. But I am definitely interested in working on this in future.

nimakarimipour: Correct, it is unrelated and also exactly correct about the explanation why this does not change within iterations.

lazaroclapp (approved):
LGTM! 

No change in test behavior is expected, and the integration test cases in [DownstreamAnalysisTest](https://github.com/nimakarimipour/NullAwayAnnotator/blob/master/core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java) (which build the code inside [downstream-dependency-test](https://github.com/nimakarimipour/NullAwayAnnotator/tree/05445b88dbe6bc8217a8297f8499ef458294d7b9/core/src/test/resources/templates/downstream-dependency-test)) provide some coverage for the `DownStreamDependencyExplorer` (cc: @ketkarameya )

ketkarameya (approved):
lgtm.

PR #44, opened 2022-08-15
Force maven-publish plugin to use shadowJar outputs.

### LINKED ISSUE #37: Install Core jar in Maven Local.
**Is your feature request related to a problem? Please describe.**
We can create the jar for `core` module via command `./gradlew shadowJar`, however, this file will be created in `build/libs/core.jar`. Most users will need to access this in maven local in the correct path according to maven publish info.

**Describe the solution you'd like**
A single task which recreates the `core` jar and installs it in the maven local.

**Describe alternatives you've considered**
Not-Applicable.

**Additional context**
Not-Applicable.

nimakarimipour:
This PR resolves #37 by forcing forcing `maven-publish` plugin to use `shadowJar` outputs.


core/build.gradle
```diff
@@ -47,6 +48,17 @@ dependencies {
 task installScanner(type: Exec) {
 
     executable "sh"
-    args "-c", "./../gradlew :scanner:publishToMavenLocal"
+    args "-c", "./../gradlew :scanner:publishToMavenLocal --rerun-tasks"
 }
 tasks.test.dependsOn("installScanner")
```
msridhar, line 53: Why do you need the `installScanner` task?  Can't you just do `tasks.test.dependsOn(':scanner:publishToMavenLocal')`?

nimakarimipour: 21af752c1c68d405fc599d54d1b6c572ce5ac2fd

Sure that works, thank you for the suggestion.

lazaroclapp (approved):
This makes sense to me. Maybe worth a quick check from @msridhar, for gradle expertise? 🙂

msridhar:
I don't see any red flags here 🙂 If it has the desired output then I think it's fine.

nimakarimipour:
> This makes sense to me. Maybe worth a quick check from @msridhar, for gradle expertise? 🙂

@lazaroclapp I'll also do a quick local run and let you know.

nimakarimipour:
> > This makes sense to me. Maybe worth a quick check from @msridhar, for gradle expertise? 🙂
> 
> @lazaroclapp I'll also do a quick local run and let you know.

@lazaroclapp @msridhar Did a local run and everything is working correctly except that it had the `-all` as a postfix to the produced (core) jar file. In order to make it convenient and adapt to the scripts, I removed it. 1dc6367e5706099220a77c54fd17d7e11a078e94

msridhar:
> @lazaroclapp @msridhar Did a local run and everything is working correctly except that it had the `-all` as a postfix to the produced (core) jar file. In order to make it convenient and adapt to the scripts, I removed it. [1dc6367](https://github.com/nimakarimipour/NullAwayAnnotator/commit/1dc6367e5706099220a77c54fd17d7e11a078e94)

The only issue with this is that then the output of the shadow task overwrites the output of the standard `jar` task, which messes up caching of Gradle results.  This only matters if you care about incremental build speed and not re-running tasks unnecessarily.  If so, this is a hack I've used to work around the issue:

https://github.com/uber/NullAway/blob/e1cae62b2a3f618ebd9b37d2f4b7fa75820e3d90/jar-infer/jar-infer-cli/build.gradle#L30-L33

nimakarimipour:
> > @lazaroclapp @msridhar Did a local run and everything is working correctly except that it had the `-all` as a postfix to the produced (core) jar file. In order to make it convenient and adapt to the scripts, I removed it. [1dc6367](https://github.com/nimakarimipour/NullAwayAnnotator/commit/1dc6367e5706099220a77c54fd17d7e11a078e94)
> 
> The only issue with this is that then the output of the shadow task overwrites the output of the standard `jar` task, which messes up caching of Gradle results. This only matters if you care about incremental build speed and not re-running tasks unnecessarily. If so, this is a hack I've used to work around the issue:
> 
> https://github.com/uber/NullAway/blob/e1cae62b2a3f618ebd9b37d2f4b7fa75820e3d90/jar-infer/jar-infer-cli/build.gradle#L30-L33

Thank you very much for the suggestion, I applied it. eb9aab5c20e094867c25728f2253fd93c39ee62c

PR #46, opened 2022-08-23
Improve computation of lower bound of number of errors on downstream dependencies in fix trees

### LINKED ISSUE #45: Incorrect computation of lower bound of number of errors on downstream dependencies in fix trees.
**Describe the bug**
At the current state, `Annotator` only considers the effect of root of the fix tree on downstream dependencies when analyzing the effect of fix tree on downstream dependencies. This is incorrect and can be much lower than the actual computable lower bound. This can lead to injecting `@Nullable` on public methods that introduces a lot of errors as a result of incorrect lower bound computation.

**To Reproduce**
Consider a fix tree with root fix `method:bar` that has `method:foo` in its chain and `bar` and `foo` leads to 0 and 100 errors on downstream dependencies if annotated as `@Nullable`. At this state the lower bound computation will only consider the root tree which is `0` and is wrong. The actual lower bound is `100`.

**Expected behavior**
When computing the lower bound of the number of errors in a fix tree, return the maximum number of errors each method can make on downstream dependencies.

**Screenshots**
Not-Applicable.

- OS: OS X
- Version 12.5
- Annotator Version: `v1.3.3-SNAPSHOT`

**Additional context**
Not-Applicable.

nimakarimipour:
This PR resolves #45 by improving computing of the lower bound of number of errors on downstream dependencies in fix trees. 

With this change, annotator will consider the maximum of number of errors triggered by each node in the fix tree as the lower bound which. Previous approach only considers the effect of the root of tree which can be really far from the the real number. This PR improves that.

### DIFF
--- a/core/src/main/java/edu/ucr/cs/riple/core/explorers/DownStreamDependencyExplorer.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/explorers/DownStreamDependencyExplorer.java
@@ -42,6 +42,8 @@ import edu.ucr.cs.riple.core.util.Utility;
 import edu.ucr.cs.riple.injector.changes.AddAnnotation;
 import edu.ucr.cs.riple.injector.location.OnMethod;
 import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.Set;
 
 /**
  * Analyzer for downstream dependencies.
@@ -149,7 +151,7 @@ public class DownStreamDependencyExplorer {
    * @param fix Fix targeting an element in target.
    * @return Effect on downstream dependencies.
    */
-  public int effectOnDownstreamDependencies(Fix fix) {
+  private int effectOnDownstreamDependencies(Fix fix) {
     if (!fix.isOnMethod()) {
       return 0;
     }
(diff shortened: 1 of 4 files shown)


core/src/main/java/edu/ucr/cs/riple/core/explorers/DownStreamDependencyExplorer.java
```diff
@@ -162,6 +164,26 @@ public int effectOnDownstreamDependencies(Fix fix) {
...
 
+  /**
+   * Returns the lower bound of number of errors of applying a fix and its associated chain of fixes
+   * on the target on downstream dependencies.
+   *
+   * @param root Root of the fix tree.
+   * @param chain Chain of the fix tree associated to root.
+   * @return Lower bound of number of errors on downstream dependencies.
+   */
+  public int computeLowerBoundOfNumberOfErrors(Fix root, Set<Fix> chain) {
```
lazaroclapp, line 175: Can we also compute the upper bound (sum) and record both values somewhere, even if we don't use the upper bound to compute the full cost right now? I kinda want to know how close to real effect on downstream deps this is, and the easiest way to gain some certainty that is close would be to see that the lower and upper bound are usually close.

lazaroclapp (approved):
LGTM. But either here or on a follow up, we should have and record the upper bound too 😃

core/src/main/java/edu/ucr/cs/riple/core/explorers/DownStreamDependencyExplorer.java
nimakarimipour: da4be8dd8867bad27678eb9b090135370502cca0

Sounds great :), implemented in this PR.

lazaroclapp: Took a quick look at the additional commit, LGTM :)

PR #51, opened 2022-08-29
Fix #50 by removing hard coded values and using environment variables

### LINKED ISSUE #50: Remove hard code values of NullAway/Scanner versions in Unit tests.
**Describe the task**
`NullAway`/`Type Annotator Scanner` versions are hardcoded in unit tests and every change in these numbers, requires modifying multiple files. This structure is extremely prone to error in updating all files.

**Expected behavior**
All test configurations must be updated according to a unique source.

**Additional context**
Not-Applicable

nimakarimipour:
This PR resolves #50 by removing all hardcoded values (`NullAway` \ `TypeAnnotatorScanner` version values) and uses environment variables to set up test project templates configurations with the current development version.

### DIFF
--- a/core/src/test/resources/templates/downstream-dependency-test/DepA/build.gradle
+++ b/core/src/test/resources/templates/downstream-dependency-test/DepA/build.gradle
@@ -42,7 +42,7 @@ dependencies {
     compileOnly project(":Target")
 
     annotationProcessor "com.uber.nullaway:nullaway:" + System.getenv('NULLAWAY_TEST_VERSION')
-    annotationProcessor "edu.ucr.cs.riple.nullawayannotator:type-annotator-scanner:" + System.getenv('TYPE_ANNOTATOR_SCANNER')
+    annotationProcessor "edu.ucr.cs.riple.nullawayannotator:type-annotator-scanner:" + System.getenv('TYPE_ANNOTATOR_SCANNER_VERSION')
     annotationProcessor files(libraryloader)
 
     compileOnly "com.google.code.findbugs:jsr305:3.0.2"
--- a/core/src/test/resources/templates/downstream-dependency-test/DepA/build.gradle
+++ b/core/src/test/resources/templates/downstream-dependency-test/DepA/build.gradle
@@ -41,8 +41,8 @@ repositories {
 dependencies {
     compileOnly project(":Target")
 
-    annotationProcessor "com.uber.nullaway:nullaway:0.9.9"
-    annotationProcessor "edu.ucr.cs.riple.nullawayannotator:type-annotator-scanner:1.3.3-SNAPSHOT"
+    annotationProcessor "com.uber.nullaway:nullaway:" + System.getenv('NULLAWAY_TEST_VERSION')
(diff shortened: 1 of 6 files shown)


core/src/test/resources/templates/downstream-dependency-test/DepA/build.gradle
```diff
@@ -41,8 +41,8 @@ repositories {
 dependencies {
     compileOnly project(":Target")
 
-    annotationProcessor "com.uber.nullaway:nullaway:0.9.9"
-    annotationProcessor "edu.ucr.cs.riple.nullawayannotator:type-annotator-scanner:1.3.3-SNAPSHOT"
+    annotationProcessor "com.uber.nullaway:nullaway:" + System.getenv('NULLAWAY_TEST_VERSION')
```
lazaroclapp, line 44: Why this rather than just defining a `deps.build.nullaway` in some `dependencies.gradle` and referencing that?

nimakarimipour: Because these project templates are standalone projects created in the temporary unit test directory, they are not part of any root project and not aware of the existing `dependencies.gradle`. Please tell me if you are suggesting a different approach than what I just said and missed something, thank you.

lazaroclapp: Ah, ok, I think I understand it: there are two instances of gradle, one which runs the tests and another which is run from the current test, and the way you are passing info between the two of those is environment variables.

Personally, I am fine with this. That said, just in case there is a cleaner way I am also missing... @msridhar , any thoughts?

msridhar: I don't fully understand the `resources/templates/...` directory setup for these tests, but assuming that is necessary (and that they can't just be top-level Gradle sub-modules), passing info via environment variables seems fine to me.

lazaroclapp (approved):
Ok, given the discussion about above, this should be ready to land :)

nimakarimipour:
Great, thank you. 🚀

core/src/test/resources/templates/downstream-dependency-test/DepA/build.gradle
nimakarimipour: @msridhar tests in `core` module are designed to verify that this module is able to compute the effect of adding `@Nullable` on an element in a code snippet correctly. There are number of tests with different code structures to cover different scenarios. The way `core` module works is that it needs a command line to rebuild the module via a `subprocess` therefore I couldn’t use the `com.google.errorprone.CompilationTestHelper`. Instead, a NullAway enrolled template `gradle` project is created for each test and tests, write their input in the `src` directory. Also, Annotator modifies the source code while running on the target module, in order to preserve the test input in any execution of tests, these project templates are created in the `junit` temporary test directory as a standalone module. For these two reasons `resource/templates/...` are used to address the mentioned limitations. Please let me know if you think I can address these in a better way :), thank you very much.

msridhar: Thanks for the explanation.  I think it makes sense to test this way, particularly since we modify the source code.  We could probably add some abstractions to test this stuff without the subprocesses, but I don't think that would make the tests any better.

PR #52, opened 2022-08-29
Prepare for upcoming version of NullAway - 0.9.10

### LINKED ISSUE #47: Update to upcoming new release of NullAway 0.9.10
**Describe the task**
Update Annotator to work with the upcoming release of NullAway. In the new release of NullAway ([PR643](https://github.com/uber/NullAway/pull/643)), information of the non-null target (the element involved in the pseudo-assignment of a @Nullable expression into a @NonNull and causing the error) is serialized along the reporting errors in the `errors.tsv`. This requires the parsing methods to be updated.

**Expected behavior**
Work as expected with `NullAway 0.9.10`.

**Screenshots**
Not-Applicable.

**Additional context**
Not-Applicable.

nimakarimipour:
#### ~~This PR is Built upon GH-51~~ GH-51 is landed.

-----

This PR resolves #47 by updating `Error` instance creation according to the latest changes.

In the new release of NullAway (Specifically [PR643](https://github.com/uber/NullAway/pull/643)), information of the non-null target (the element involved in the pseudo-assignment of a @Nullable expression into a @NonNull and causing the error) is serialized along the reporting errors in the `errors.tsv`.

Header of `errors.tsv` files is changed from below:
```
MESSAGE_TYPE    MESSAGE   CLASS   METHOD
```
to:
```
message_type    message    enc_class    enc_method    target_kind    target_class   target_method     param    index    uri    
```

This requires updates on how `Error` instances are created due to change in `errors.tsv` files contents.

-----
~~Note: tests for `core` module are temporarily excluded in this PR for CI as they are only compatible with NullAway verion `0.9.10`.~~ _All_ tests are included since `NullAway 0.9.10` is released.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -49,7 +49,7 @@ tasks.test.dependsOn(':type-annotator-scanner:publishToMavenLocal')
 
 // Set up environment variables for test configuration tu run with the latest development version.
 tasks.test.doFirst {
-    environment "NULLAWAY_TEST_VERSION", "0.9.9"
+    environment "NULLAWAY_TEST_VERSION", "0.9.10-SNAPSHOT"
     environment "TYPE_ANNOTATOR_SCANNER_VERSION", project.version
 }
 
@@ -72,4 +72,17 @@ jar {
 // To remove the "-all" postfix from the produced jar file.
 shadowJar {
     archiveClassifier = null
+}
+
+
+
+// TODO: Remove below once NullAway 0.9.10 is released
+tasks.test {
+    filter {
+        //exclude all tests below as they are only compatible with NullAway 0.9.10 (The upcoming version).
+        excludeTestsMatching "edu.ucr.cs.riple.core.CoreTest"
+        excludeTestsMatching "edu.ucr.cs.riple.core.DeepTest"
+        excludeTestsMatching "edu.ucr.cs.riple.core.InheritanceTest"
+        excludeTestsMatching "edu.ucr.cs.riple.core.DownstreamAnalysisTest"
+    }
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -47,6 +47,12 @@ dependencies {
 
 tasks.test.dependsOn(':type-annotator-scanner:publishToMavenLocal')
 
+// Set up environment variables for test configuration tu run with the latest development version.
+tasks.test.doFirst {
+    environment "NULLAWAY_TEST_VERSION", "0.9.9"
+    environment "TYPE_ANNOTATOR_SCANNER_VERSION", project.version
+}
+
 publishing {
     publications {
         shadow(MavenPublication) { publication ->
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -49,7 +49,7 @@ tasks.test.dependsOn(':type-annotator-scanner:publishToMavenLocal')
 
 // Set up environment variables for test configuration tu run with the latest development version.
 tasks.test.doFirst {
-    environment "NULLAWAY_TEST_VERSION", "0.9.10-SNAPSHOT"
+    environment "NULLAWAY_TEST_VERSION", "0.9.10"
     environment "TYPE_ANNOTATOR_SCANNER_VERSION", project.version
 }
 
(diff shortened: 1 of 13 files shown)


core/build.gradle
```diff
@@ -66,4 +72,17 @@ jar {
 // To remove the "-all" postfix from the produced jar file.
 shadowJar {
     archiveClassifier = null
-}
\ No newline at end of file
+}
+
+
+
+// TODO: Remove below once NullAway 0.9.10 is released
```
lazaroclapp, line 79: Are we blocked on a NullAway 0.9.10 release here? I think it might be a bad idea to disable tests from CI, rather than holding this PR until NullAway 0.9.10 is out. But if this is a blocker for you, let me know and we can speed up the upstream release!

nimakarimipour: Yes, all tests in `core` module except `ConfigurationTest` requires `NullAway 0.9.10`. One other approach is that I can temporarily update the CI to clone and install `0.9.10-SNAPSHOT`. Please let me know if you prefer this approach. Thank you.

nimakarimipour: Brought back all tests since `NullAway 0.9.10` is released. bc5fe129bcb9079dff981357864167969b3f1df2
This is ready for review now.

core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/ConflictGraph.java
```diff
@@ -74,7 +74,7 @@ public Node addNodeToVertices(Fix fix) {
    * in the same group. A greedy algorithm is used to find the solution.
    */
   // TODO: Remove SuppressWarnings below later.
-  @SuppressWarnings("JdkObsolete")
+  @SuppressWarnings({"JdkObsolete"})
```
lazaroclapp, line 77: Left-over change? Curly brackets shouldn't be needed if only one checker is being suppressed. Though I guess this works either way :)

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -33,20 +36,36 @@ public class Error extends Enclosed {
...
+      String encClass,
+      String encMethod,
+      @Nullable Location nonnullTargetLocation) {
     super(encClass, encMethod);
     this.messageType = messageType;
     this.message = message;
+    this.location = nonnullTargetLocation;
   }
 
   @Override
```
lazaroclapp, line 71: Are `equals(...)` and `hashCode(...)` out of date here? They don't even seem to take `message` into account, just `messageType`, so it feels like these would often clash for non-equivalent `Error` objects. (These get used when comparing two `Bank<Error>` collections, so not sure if we are missing stuff there, maybe it is fine if we are only counting the errors of the same type before and after, but then we should use something other than `.equals(...)` since we aren't really checking for equality here)

lazaroclapp (commented):
A couple questions, mostly about `Error.equals` and `Error.hashCode` in light of the change to the object's fields...

core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/ConflictGraph.java
nimakarimipour: Yes this is a left-over, thank you for noticing this, removed it. c82365feb4aef9a2988f3f9a773ad20afe55d2e4

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
nimakarimipour: Yes, they are updated thanks for the note, the reason that it was working correctly so far, is that we are only looking at the number of differences in number of errors, and they weren’t actually checked if they are equal. This hash function was there just for `Bank` to be able to compute a hash value for an `Error` instance. 

We weren’t even processing `Message Type` since we determine if an error is resolvable by checking if a new fix is triggered.

Both methods are updated now corresponding to the added fields, both to avoid clashing and in GH-53 we do need to see the difference in errors contents.

core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -77,11 +77,14 @@ public boolean equals(Object o) {
       return false;
     }
     Error error = (Error) o;
-    return messageType.equals(error.messageType);
+    return messageType.equals(error.messageType)
+        && message.equals(error.message)
+        // Since location is @Nullable, used Objects.equal.
```
lazaroclapp, line 82: This is no longer called location, so this comment is outdated.

lazaroclapp (approved):
Small nit below, but otherwise this looks ready to land to me once that's resolved.

PR #54, opened 2022-08-30
Rename MethodInheritanceTree to MethodDeclarationTree

nimakarimipour:
#### ~~This PR is built upon GH-52~~ GH-52 is landed.

This PR simply renames `MethodInheritanceTree` to `MethodDeclarationTree`. 

`MethodInheritanceTree` has a top down tree structure of method declarations where each method is child of their immediate overriding method. Although the structure of the tree is based on the inheritance, `MethodDeclarationTree` is a better name as in many use cases, it's api is only called to locate a method declaration and use its containing data.


lazaroclapp (approved):
Rename sounds good to me!

nimakarimipour:
@lazaroclapp Thank you for the review, will land it now.

PR #63, opened 2022-09-04
Fix #62 by adding configuration mode for downstream dependency analysis.

### LINKED ISSUE #62: Prevent build failures on downstream dependencies due to changes in upstream
**Is your feature request related to a problem? Please describe.**

If annotator processes the target module with `--downstream-dependency-analysis-activated` flag, it will collects effects of making public methods `@Nullable` on downstream dependencies and considers the overall effect while making inference decisions. Some annotations can reduce the number of errors locally far more than the number of triggered errors on downstream dependencies and can have a positive overall impact, hence, these  annotations will get injected to the target module (even though this will lead to build errors on downstream dependencies). 

It is a common use case where the user would like to run the AutoAnnotator on `target` module with following constraints:
- No modification on downstream dependencies is allowed.
- No build failures on downstream dependencies after enrolling the target module.

**Describe the solution you'd like**

A configuration mode which if enabled, it guarantees that no changes in upstream will trigger an error on downstream dependencies. This should also cover detection of flow of `@Nullable` back to upstream. See example below:

#### Target Module:
```java
class Foo{

     public Object returnNullBad(){
              // Assume there are 5 branches ending with return null;
              return null;
     }

     public Object returnNullGood(){
              // Assume there are 5 branches ending with return null;
              return null;
     }

     public void takeNullGood(Object param){ // Called only in downstream dependencies.
              if(param != null){
                     param.deref();
              }
     }

     public void takeNullBad(Object){ // Called only in downstream dependencies.
             param.deref();
     }
}
```

#### Downstream Dependency Module:
```java
class Dep{

     Object field = new Object();
     Foo foo = new Foo();
     
     public void bar(){
             field = foo.returnNullBad();
             foo.takeNullGood(foo.returnNullBad());
             foo.takeNullGood(foo.returnNullGood());
             foo.takeNullBad(foo.returnNullGood());
     }
}
```

If Annotator only focus on the overall impact, it will annotate the `Target` module as below:
```java
class Foo{

     @Nullable   // This will resolve 5 errors locally and triggers 1 unresolvable and 1 resolvable error (on Dep) overall: -4
     public Object returnNullBad(){
              // Assume there are 5 branches ending with return null;
              return null;
     }

     @Nullable // This will resolve 5 errors locally and 1 resolvable error (change on Target) overall: -5
     public Object returnNullGood(){
              // Assume there are 5 branches ending with return null;
              return null;
     }

     // This will resolve 1 error on downstream dependency and create no error locally
     public void takeNullGood(@Nullable Object param){ // Called only in downstream dependencies.
              if(param != null){
                     param.deref();
              }
     }

     // Since @Nullable injection will create an error locally and the overall is +1,
     // it will not get injected, however, this will leave the error on downstream unresolved.
     public void takeNullBad(Object){ // Called only in downstream dependencies.
             param.deref();
     }
}
```

With the requested configuration mode, annotator should annotate `Target` module as below:
```java
class Foo{

     @NullUnmarked //To resolve the error on downstream dependency.
     public Object returnNullBad(){
              // Assume there are 5 branches ending with return null;
              return null;
     }

     @Nullable
     public Object returnNullGood(){
              // Assume there are 5 branches ending with return null;
              return null;
     }

     public void takeNullGood(@Nullable Object param){ // Called only in downstream dependencies.
              if(param != null){
                     param.deref();
              }
     }

     @NullUnmarked //To resolve the error on downstream dependency.
     public void takeNullBad(Object){ // Called only in downstream dependencies.
             param.deref();
     }
}
```

**Describe alternatives you've considered**
None.

**Additional context**
Not-Applicable.

nimakarimipour:
#### ~~This PR is build upon GH-64~~. GH-64 is landed now.
----

This Resolves #62 by adding a configuration mode `Strict`.

With this PR, we can now configure the analysis mode which will impact inference decisions.

We have four different modes of analysis:
1. `Local`: Only considers the local effect of fixes.
2. `Lower Bound`: Considers the lower bounds of number of errors on downstream dependencies.
3. `Upper Bound`: Considers the upper bound of number of errors on downstream dependencies.
4.  `Strict`: Guarantees that all errors in downstream due to changes in upstream will be resolved.

To set the above modes via command line, pass `-am` or `--analysis-mode` followed by one of the options below:
- `default`
- `upper_bound`
- `lower_bound`
- `strict`

Or set one of the values above In the `.json` config file at:
```json
"DOWNSTREAM_DEPENDENCY_ANALYSIS":{
     "ANALYSIS_MODE": "mode"
}
```


nimakarimipour:
Hi @lazaroclapp, this PR is ready for review however, it needs GH-64 to be landed first, will resolve your comments on that now :) Thank you.

core/src/test/java/edu/ucr/cs/riple/core/AnalysisModeTest.java
```diff
@@ -0,0 +1,88 @@
...
+        .addExpectedReports(
+            // Resolves 6 errors locally and all errors on downstream dependencies can be resolved
+            // with no new triggered error, therefore the effect is -6. The fix triggers no error
+            // in downstream dependencies, should be approved.
+            new TReport(
+                new OnMethod("Foo.java", "test.target.Foo", "returnNullGood()"), -6, APPROVE),
+            // Resolves 6 errors locally but creates 1 error on downstream dependencies that cannot
+            // be resolved, therefore the effect is -5. The fix triggers 1 error in downstream
+            // dependencies, should be rejected.
+            new TReport(new OnMethod("Foo.java", "test.target.Foo", "returnNullBad()"), -5, REJECT))
```
lazaroclapp, line 56: 👍

core/src/test/java/edu/ucr/cs/riple/core/AnalysisModeTest.java
```diff
@@ -0,0 +1,88 @@
...
+                expected.root.equals(found.root)
+                    && expected.getExpectedValue()
+                        == found.getOverallEffect(coreTestHelper.getConfig())
+                    && Objects.equals(expected.getTag(), found.getTag()))
+        .toDepth(5)
+        .disableBailOut()
+        .enableDownstreamDependencyAnalysis(LOWER_BOUND)
+        .start();
+  }
+}
```
lazaroclapp, line 88: Do we want an upper bound test? (That will need a second dependency, but it's probably a good idea to check that lower/upper bound do what they are supposed to do when there is more than one dependency)

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
```diff
@@ -205,6 +206,41 @@ public void testConfigFilesHaveDifferentUUID() {
     }
   }
 
+  @Test
+  public void testAnalysisModeFlags() {
+    List<CLIFlag> baseFlags = new ArrayList<>(requiredFlagsCli);
+    Config config;
+    // Check mode downstream dependency off.
+    config = new Config(makeCommandLineArguments(requiredFlagsCli));
```
lazaroclapp, line 214: Should this be taking `baseFlags` here, since we already defined it? Otherwise I'd expect the definition of `baseFlags` to happen below this line (and after the following assertEquals)

core/src/main/java/edu/ucr/cs/riple/core/global/GlobalAnalyzer.java
```diff
@@ -88,4 +88,14 @@ public interface GlobalAnalyzer {
...
+
+  /**
+   * Checks if fix triggers any unresolvable error in downstream dependencies. Unresolvable errors
+   * are errors that either no annotation can resolve them, or the corresponding fix is targeting an
+   * element outside the target module.
+   *
+   * @param fix Fix to apply
+   * @return true if the method triggers an unresolvable error in downstream dependencies.
+   */
+  boolean isDestructiveFix(Fix fix);
```
lazaroclapp, line 100: I get the point of this method, but why "destructive"? Why not simply `isDownstreamFix(...)` or `isFixForcingDownstreamChanges(...)`? Also, is this actually checking for errors without a fix? Documentation says "either no annotation can resolve them or [not in the target module]" but the only checks I see below are for the second case.

core/src/main/java/edu/ucr/cs/riple/core/AnalysisMode.java
```diff
@@ -0,0 +1,158 @@
...
+   * @param downStreamDependenciesAnalysisActivated if true, downstream dependency analysis is
+   *     activated.
+   * @param mode passed mode.
+   * @param useDefault if true, no value is provided by the user and the default mode will be
+   *     returned. (If downstream dependency analysis is activated default is {@link
+   *     AnalysisMode#LOWER_BOUND}, otherwise the default is {@link AnalysisMode#LOCAL}).
+   * @return the corresponding {@link AnalysisMode}.
+   */
+  public static AnalysisMode parseMode(
+      boolean downStreamDependenciesAnalysisActivated, @Nullable String mode, boolean useDefault) {
```
lazaroclapp, line 136: Why do we need `useDefault` here or a `@Nullable` `mode`, if simply passing `mode = "default"` will have the same effect. Looks like the logic and interface of this method can be greatly simplified.

core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -330,6 +344,11 @@ public Config(String[] args) {
     this.exhaustiveSearch = cmd.hasOption(exhaustiveSearchOption.getLongOpt());
     this.downStreamDependenciesAnalysisActivated =
         cmd.hasOption(downstreamDependenciesActivationOption.getLongOpt());
+    this.mode =
+        AnalysisMode.parseMode(
+            this.downStreamDependenciesAnalysisActivated,
+            cmd.getOptionValue(analysisMode),
+            !cmd.hasOption(analysisMode));
```
lazaroclapp, line 351: Could simplify `parseMode` as described above and then here do:

```
    this.mode =
        AnalysisMode.parseMode(
            this.downStreamDependenciesAnalysisActivated,
            cmd.getOptionValue(analysisMode, "default"));
```

No need for the third argument or for supporting `null` modes.

core/src/main/java/edu/ucr/cs/riple/core/AnalysisMode.java
```diff
@@ -0,0 +1,158 @@
...
+  STRICT {
+    @Override
+    public void tag(Config config, GlobalAnalyzer analyzer, Collection<Report> reports) {
+      reports.forEach(
+          report -> {
+            // Check for destructive methods.
+            if (report.containsDestructiveMethod(analyzer)) {
+              report.tag(REJECT);
+              return;
+            }
```
lazaroclapp, line 67: Worth a `Preconditions.checkArgument(report.getUpperBoundEffectOnDownstreamDependencies() == 0)` sanity check here?

lazaroclapp (commented):
Some suggestions below and a question about testing, but overall this looks good to me!

core/src/test/java/edu/ucr/cs/riple/core/ConfigurationTest.java
nimakarimipour: Updated it 7569c5ed1b34905c7dcaaf96668fc7f24beed225

core/src/main/java/edu/ucr/cs/riple/core/global/GlobalAnalyzer.java
nimakarimipour: f49a031dc5f32ee6d7792a15a869f57a2c2aa568 renamed it to `isFixForcingDownstreamChanges`.
It actually checks if the triggered error is not resolvable by any annotation. Triggered errors that no annotation can resolve them have a `nonnull` target with all values equal to `null` which `declaredInModule` returns `false` for them.

core/src/main/java/edu/ucr/cs/riple/core/Config.java
nimakarimipour: Simplified 56d8f5ed2ed954686f172272e79b425def7a30ee Thank you very much.

core/src/main/java/edu/ucr/cs/riple/core/AnalysisMode.java
nimakarimipour: Added fc76915943c76a462169fba5fe5d7f86489ce8dc.

nimakarimipour:
@lazaroclapp I resolved all the comment, this is ready for another review. Thank you 🙂 .

core/src/main/java/edu/ucr/cs/riple/core/global/GlobalAnalyzer.java
lazaroclapp: > It actually checks if the triggered error is not resolvable by any annotation. Triggered errors that no annotation can resolve them have a nonnull target with all values equal to null which declaredInModule returns false for them.

Ok, then maybe `isNotFixableOnTarget(...)` might be a better name. Sorry for the back and forth here :)

Might be worth a comment on that implicit check for why this caches fully non-resolvable errors too.

lazaroclapp (commented):
One question below about lower/upper bound, sorry I didn't catch it sooner, but the test doesn't add up to my mental model of what I thought those two methods were computing (the max and min errors fixed for a given chain of fixes across a series of downstream dependencies, where the effect on a single dependency was known).

Also, please note the nit about the name of `isFixForcingDownstreamChanges` -> `isNotFixableOnTarget`, since that's in the thread for one of the previous round of comments 🙂

core/src/main/java/edu/ucr/cs/riple/core/global/GlobalAnalyzer.java
nimakarimipour: Sure, `isNotFixableOnTarget`  sound great.
Added a comment and renamed it. e9da6a7b76c33f6b6742c99d48131e9dbf910017

core/src/main/java/edu/ucr/cs/riple/core/AnalysisMode.java
```diff
@@ -133,11 +135,11 @@ public void tag(Config config, GlobalAnalyzer analyzer, Collection<Report> repor
    * @return the corresponding {@link AnalysisMode}.
    */
   public static AnalysisMode parseMode(
-      boolean downStreamDependenciesAnalysisActivated, @Nullable String mode, boolean useDefault) {
+      boolean downStreamDependenciesAnalysisActivated, String mode) {
     if (!downStreamDependenciesAnalysisActivated) {
       return LOCAL;
     }
-    if (useDefault || mode == null) {
+    if (mode == null) {
```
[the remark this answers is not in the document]
nimakarimipour: 770d4c96d2b58b25d44bc17cc10e5979481f14ad

core/src/test/java/edu/ucr/cs/riple/core/AnalysisModeTest.java
```diff
@@ -85,4 +86,27 @@ public void lowerBoundModeTest() {
...
+
+  @Test
+  public void upperBoundModeTest() {
+    coreTestHelper
+        .addExpectedReports(
+            new TReport(
+                new OnMethod("Foo.java", "test.target.Foo", "returnNullGood()"), -6, APPROVE),
+            // Resolves 6 errors locally but creates 1 error on downstream dependencies that cannot
+            // be resolved, one of the triggered fixes in the tree, triggers an error in downstream
+            // dependency, the overall effect is -6 + 1 + 1 = -4 in upper bound mode.
```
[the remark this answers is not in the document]
nimakarimipour: Regarding upper bound and lower bound, please consider the assumptions below:
1. A root fix resolving error is `rf`.
2. `rf` triggers `f0`, `f1` and `f2` all targeting elements in the target module.
3. If `rf`,`f0`, `f1`, `f2` applied, it will decrease the number of errors on target by 6 -> effect = -6.
And below is the result of analysis of downstream dependencies on `dep1`, `dep2` and `dep3`:

`Cell[i][j]` is the number of triggered errors for fix `i` on dependency `j`.

|Fix\Dep|`dep1`|`dep2`|`dep3`| total |
|:------:|:-----:|:------:|:------:|:-----:|
|   `rf`   |   0    |      1     |      2    |     3   |
|  ` f0`  |  1     |      1     |      2    |     4   |
|   `f1`   |   1     |      9    |      0    |     10 |
|   `f2`  |   0    |      0    |      1     |     1   |

With the current approach, regardless of the mode configuration we can detect the set of triggered errors for each fix on each module individually.

For chain `rf` -> `f0`, `f1`, `f2`, since all fixes are applied we can't measure the effectiveness accurately since they may trigger the same error. 
In `upper bound` mode we assume all errors are distinct therefore the overall effect of the chains is:
-6 + 3 + 4 + 10 + 1 = 12.

In `lower bound` mode, we just consider the maximum number of errors as the effect of the whole chain on downstream dependencies:
-6 + 10 = 4.

Therefore upper bound and lower bound is regardless of the number of downstream dependencies.


And the part you mentioned regarding that filtering, that only filters triggered errors that are resolved due to a fix in the tree, it does not detect identical errors that two fixes in a tree can trigger. Please let me know your thoughts on this. Thank you.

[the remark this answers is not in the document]
lazaroclapp: I had to think a bit to parse this explanation, but I think it tracks.

Let me see if I got it...

First, for a single fix (not a tree of fixes, just the individual `rf` or `f{i}` in isolation), there is no difficulty in computing the _precise_ effect of the fix on each dependency or the _precise_ effect of the fix across all dependencies (the **total** column above). The difficulty lies in combining the effects of all fixes in the tree, and this is true whether there are multiple dependencies or a single one. Correct?

So, going by the example from the test case, we have one dep and three fixes:

- `rf` : "Add `@Nullable` to `returnNullBad()`"
- `f0` : "Add `@Nullable` to argument 1 of `takeNullGood(...)`" (because of `bar()` in `Dep`)
- `f1` : "Add `@Nullable` to `getNull()`" (because that method returns the result of `returnNullBad()`)

So the table looks like:

|Fix\Dep|`dep`|explanation|
|:------:|:-----:|:-----:|
|   `rf`   |   1    |  2 errors generate, but one resolved by f1 |
|  ` f0`  |  0     |  |
|   `f1`   |   1     | body of `getNullDown()` |

Right?

And the effect on the target itself is removing 5 errors from `returnNullBad()` and one from `getNull()`, so -6+2=-4, which is both correct and what the "upper bound" method gives here.

Meanwhile, "lower bound" gives -6+1=-5, because it can't tell if the new error that shows when applying `f1` is a repeat of the one that shows up when applying `rf`.

And this:

> And the part you mentioned regarding that filtering, that only filters triggered errors that are resolved due to a fix in the tree, it does not detect identical errors that two fixes in a tree can trigger.

Is just the logic that filters the error from `foo.takeNullGood(foo.returnNullBad());` in `Dep::bar()`, because it can see a fix for that specific error in the tree (namely, it sees `f0`).

lazaroclapp (approved):
Assuming I got the explanation right in the long comment above, this LGTM! 😁

core/src/test/java/edu/ucr/cs/riple/core/AnalysisModeTest.java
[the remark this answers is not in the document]
nimakarimipour: Yes that is absolutely correct. 👍🏻

nimakarimipour:
> Assuming I got the explanation right in the long comment above, this LGTM! 😁

Yes that is correct, I will land this now, thank you for the review 😄.

PR #68, opened 2022-09-08
Fix #56 by introducing reports cache and reruning analysis at the end of the iteration

### LINKED ISSUE #56: Rerun analysis at the end of the process.
**Describe the task**
Annotator caches the result of explorations and will reuse them in all further iterations. If a fix tree is not reducing the number of errors, it will be tagged as an `unuseful` branch and annotator will never investigate its effect in further iterations. However, at the end of the process, the state of the program has altered due to injection of series of annotations in previous iterations and those tagged branches can be potentially `usefull` and reduce the number of errors further. Annotator should invalidate the branches tags at the end of the process and rerun the analysis for __one__ more iteration.

**Expected behavior**
Invalidate reports cache and rerun analysis for only  __one__ more iteration.

**Additional context**
Not-Applicable.

nimakarimipour:
#### ~~This PR is build upon GH-62.~~ GH-62 is landed.

This PR resolves #56 by introducing `ReportCache` class and breaking down `explore` method in annotator. With this PR, at the analysis it runs reruns the `inference/injection` phase without exclusion of already processed `fixes`.


core/src/main/java/edu/ucr/cs/riple/core/ReportCache.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Updates the state of cache with the latest processed reports.
+   *
+   * @param reports Set of the latest processed reports.
+   */
+  public void update(ImmutableSet<Report> reports) {
+    int size = store.keySet().size();
+    reports.forEach(
+        report -> {
+          store.putIfAbsent(report.root, report);
```
lazaroclapp, line 78: What's the reason here to call `putIfAbsent` followed by modifying the existing report to match the fields of the new report?

i.e. why isn't this just:

```
reports.forEach( report -> store.put(report.root, report));
```

?

Not saying there isn't a reason, but, if there is, then it might be worth a comment explaining it :)

lazaroclapp (commented):
Overall looks good, but I do have a quick question...

core/src/main/java/edu/ucr/cs/riple/core/ReportCache.java
nimakarimipour: This is probably a left over from previous designs. This piece has been with us from the start :)))

I don't immediately see a reason to keep this with our latest design. Please Let me think about it a little more and I will tell you if we can safely remove it.

lazaroclapp (approved):

core/src/main/java/edu/ucr/cs/riple/core/ReportCache.java
nimakarimipour: Hi @lazaroclapp I don't think this has any side effects, I changed it to using the implementation you suggested. Thank you.

nimakarimipour:
@lazaroclapp Thank you for the review, will land this.

Issue #21, opened 2022-07-15
Add Error Prone checkers and Enroll all modules in NullAway


nimakarimipour:
**Describe the task**
Add `Error Prone` checkers / `NullAway` to this project.

**Expected behavior**
Not-Applicable.

**Screenshots**
Not-Applicable.

**Additional context**
Not-Applicable.

nimakarimipour:
Update: Error Prone checkers are added through GH-41. Will close this issue once the repo is enrolled in `NullAway`

PR #69, opened 2022-09-08
Fix #55 by supporting @NullUnmarked

### LINKED ISSUE #55: Support @NullUnmarked
**Describe the task**
Add a configuration mode that if enabled, at the end of the process, annotator marks the closest enclosing method of the remaining errors with `@NullUnmarked` annotation to guarantee NullAway checks pass.

**Expected behavior**
All remaining errors should be enclosed by the `@NullUnmarked` annotation.

**Additional context**
`@NullUnmarked` can be located [here](https://github.com/jspecify/jspecify/blob/main/src/main/java/org/jspecify/nullness/NullUnmarked.java).

nimakarimipour:
~~This PR is built upon GH-68.~~ GH-68 is landed.

This PR resolves #55 by adding a configuration mode which if activated, all remaining errors will be resolved with following the rules below:
1. Enclosing method of triggered errors will be marked as `@NullUnmarked`.
2. Uninitialized fields (inline or in constructor) will be annotated as `@SuppressWarnings("NullAway.Init")`.
3. Explicit `@Nullable` assignments to fields will be annotated as `@SuppressWarnings("NullAway")`.

With this PR, the target module will enroll into NullAway with no triggered errors.

To enable this mode, `cli` flag below must be passed to Annotator:
```
-fr or --force-resolve [followed by the "@NullUnmarked" fully qualified name]
```


core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -160,6 +160,20 @@ public void multiple_field_declaration_test() {
...
+  public void multipleFieldDeclarationTestForceResolveEnabled() {
+    coreTestHelper
+        .addInputDirectory("test", "multiplefielddeclration")
+        .disableBailOut()
+        .addExpectedReports(
+            new TReport(
+                new OnField("Main.java", "test.Main", newHashSet("f1", "f2", "f3", "f4")), 9),
+            new TReport(new OnField("Main.java", "test.Main", singleton("f5")), 1))
+        .toDepth(1)
+        .enableForceResolve()
```
lazaroclapp, line 173: See comment in `strictModeWithForceResolveTest()`, worth mentioning that we are not just altering the configuration, but adding extra checking here.

core/src/test/java/edu/ucr/cs/riple/core/AnalysisModeTest.java
```diff
@@ -109,4 +109,30 @@ public void upperBoundModeTest() {
...
+        .setPredicate(
+            (expected, found) ->
+                expected.root.equals(found.root)
+                    && expected.getExpectedValue()
+                        == found.getOverallEffect(coreTestHelper.getConfig())
+                    && Objects.equals(expected.getTag(), found.getTag()))
+        .toDepth(5)
+        .disableBailOut()
+        .enableDownstreamDependencyAnalysis(STRICT)
+        .enableForceResolve()
```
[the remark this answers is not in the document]
lazaroclapp: Edit: looked further into the implementation of `CoreTestHelper`, I think it's worth having a comment here along the lines of:

```
// This causes the test to report a failure if any errors remain when building the target.
```

Long term, I still think we would like to have test cases that can check the "expected output" source code for the target, including where all the annotations got placed, but obviously not for this week 🙂

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -216,4 +224,71 @@ private ImmutableSet<Report> processTriggeredFixes(
...
+  private void forceResolveRemainingErrors(
+      FieldDeclarationAnalysis fieldDeclarationAnalysis, MethodDeclarationTree tree) {
+    // Collect regions with remaining errors.
+    Utility.buildTarget(config);
+    List<Error> remainingErrors = Utility.readErrorsFromOutputDirectory(config.target);
+    Set<Fix> remainingFixes =
+        Utility.readFixesFromOutputDirectory(
+            config.target, Fix.factory(config, fieldDeclarationAnalysis));
+
+    // Collect all regions for NullUnmarked.
```
lazaroclapp, line 252: Probably worth adding here some explanation of which regions qualify for NullUnmarked. Maybe:

```
// For all errors in regions which correspond to a method's body, we can add @NullUnmarked at the method level.
```

?

p.s. presumably, in the future, it might make sense to add `@NullUnmarked` to classes (e.g. inner classes) in specific circumstances, but for now the above seems like a correct characterization. No?

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -216,4 +224,71 @@ private ImmutableSet<Report> processTriggeredFixes(
...
+                    return false;
+                  }
+                  OnField onField = fix.toField();
+                  return onField.clazz.equals(fix.encClass()) && fix.encMethod().equals("");
+                })
+            .map(fix -> new AddAnnotation(fix.toField(), "SuppressWarnings", "NullAway"))
+            .collect(Collectors.toSet());
+    injector.injectAnnotations(suppressWarningsAnnotations);
+
+    // Collect NullAway.init SuppressWarnings
```
lazaroclapp, line 280: `NullAway.Init`

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -216,4 +224,71 @@ private ImmutableSet<Report> processTriggeredFixes(
...
+    // Collect all regions for NullUnmarked.
+    Set<AddAnnotation> nullUnMarkedAnnotations =
+        remainingErrors.stream()
+            // filter non-method regions.
+            .filter(error -> !error.encMethod().equals("null"))
+            // find the corresponding method nodes.
+            .map(error -> tree.findNode(error.encMethod(), error.encClass()).location)
+            // impossible, just sanity check or future nullness checker hints
+            .filter(Objects::nonNull)
+            .map(location -> new AddAnnotation(location, config.nullUnMarkedAnnotation))
```
lazaroclapp, line 261: Don't we need a way to pass the reason this error wasn't/couldn't be fixed here somehow? (To be added in a comment, not a string, but either way it would be good to have)

lazaroclapp (commented):
Some comments. Mostly concerned on a few edge-cases around the injector and whether or not we can easily add comments instead of inner-strings for `@NullUnmarked` and comment explanations to `@SuppressWarnings` annotations too.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
```diff
@@ -0,0 +1,186 @@
...
+            "   @SuppressWarning(\"NullAway.init\")",
+            "   Object h = new Object();",
+            "   public void test(Object f) {",
+            "      h = f;",
+            "   }",
+            "}")
+        .addChanges(
+            new AddAnnotation(
+                new OnField("Super.java", "com.uber.Super", Collections.singleton("h")),
+                "SuppressWarning",
```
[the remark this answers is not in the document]
nimakarimipour: Thank you very much noticing that. They are actually got updated in the `core` module when I noticed this issue. They are also updated in tests now b9f3411b187a435af50df96b7666edfb5b32fd98.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
```diff
@@ -0,0 +1,186 @@
...
+                "NullAway.init"))
```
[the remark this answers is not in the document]
nimakarimipour: Only `NullAway.Init` works, in the `core` tests `NullAway.Init` will be injected and tested. They are also updated in tests module as well. b9f3411b187a435af50df96b7666edfb5b32fd98

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
```diff
@@ -0,0 +1,186 @@
...
+            "   Object h = new Object();",
+            "   public void test(Object f) {",
+            "      h = f;",
+            "   }",
+            "}")
+        .expectOutput(
+            "package com.uber;",
+            "public class Super {",
+            "   @SuppressWarning(\"something else\")",
+            "   @SuppressWarning(\"NullAway.init\")",
```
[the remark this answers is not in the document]
nimakarimipour: Thank you for noticing this, didn't know that. I updated the whole mechanism.
3ad79970e4952e4e194af236515fc2c512d27a64

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
```diff
@@ -0,0 +1,186 @@
...
+            "   public void test(Object f) {",
+            "      h = f;",
+            "   }",
+            "}")
+        .expectOutput(
+            "package com.uber;",
+            "import edu.ucr.NullUnmarked;",
+            "public class Super {",
+            "   Object h = new Object();",
+            "   @NullUnmarked(\"Prefix: Creates 2 errors on downstream dependencies\")",
```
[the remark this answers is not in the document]
nimakarimipour: Sure, I replaced all `@NullUnmarked` annotations in test with `edu.ucr.CustomNull` with same annotation arguments. I think regardless of `@NullUnmarked` annotation, the tests are useful.

injector/src/test/java/edu/ucr/cs/riple/injector/RemovalTest.java
```diff
@@ -118,12 +120,15 @@ public void remove_annot_param_full_name() {
                     "com.uber.Super",
                     "test(@javax.annotation.Nullable java.lang.Object)",
                     0),
-                "javax.annotation.Nullable"))
-        .start();
+                "javax.annotation.Nullable"));
+    assertThrows(AssertionError.class, () -> injectorTestHelper.start());
```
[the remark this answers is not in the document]
nimakarimipour: eef31f8b08e026b34afb55773f1c0859cf14d81b sorry for the confusion, the comment is for the test `remove_annot_param_full_name` and not `remove_annot_field`. Sorry for that eef31f8b08e026b34afb55773f1c0859cf14d81b

The intention of this test to make sure that we do not remove annotations with full names as injector does not inject any annotation with fully qualified name.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddAnnotation.java
```diff
@@ -22,21 +22,56 @@
...
+
+    if (argument == null || argument.equals("")) {
+      node.addMarkerAnnotation(annotationSimpleName);
+    } else {
+      node.addAnnotation(annotationExpr);
+    }
   }
 
   @Override
+  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
nimakarimipour: That annotation was for the whole class, but since I was modifying the class, I thought maybe it is better to refine this and not have it for the full class.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddAnnotation.java
```diff
@@ -22,21 +22,56 @@
...
+import com.github.javaparser.ast.expr.StringLiteralExpr;
+import com.github.javaparser.ast.nodeTypes.NodeWithAnnotations;
 import edu.ucr.cs.riple.injector.location.Location;
+import javax.annotation.Nullable;
 import org.json.simple.JSONObject;
 
 public class AddAnnotation extends Change {
-  public AddAnnotation(Location location, String annotation) {
+
+  /** Argument of the annotation. If null, the added annotation will be a marker annotation. */
```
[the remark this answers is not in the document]
nimakarimipour: Thank you for the description, I renamed the classes and the design is much cleaner.
Updated annotation classes to show this explicitly 84daa4e03f3d24f8078cf6895e9419c5a6074393  and ede7896c2bbb33492a5dada3f810ab36afcce999.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddAnnotation.java
```diff
@@ -22,21 +22,56 @@
...
-    return this.location.apply(tree, annotation, true);
+  public void visit(NodeWithAnnotations<?> node) {
+    NodeList<AnnotationExpr> annotations = node.getAnnotations();
+    AnnotationExpr annotationExpr =
+        this.argument == null
+            ? new MarkerAnnotationExpr(annotationSimpleName)
+            : new SingleMemberAnnotationExpr(
+                new Name(annotationSimpleName), new StringLiteralExpr(argument));
+
+    // Check if annot already exists.
```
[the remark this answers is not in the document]
nimakarimipour: I actually updated this implementation and added a whole set of new tests that covers the cases you mentioned. Please find it here 3ad79970e4952e4e194af236515fc2c512d27a64.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/Change.java
```diff
@@ -56,6 +77,7 @@ public int hashCode() {
     return Objects.hash(location, annotation);
   }
 
+  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
nimakarimipour: That annotation was for the whole class, but since I was modifying the class, I thought maybe it is better to refine this and not have it for the full class.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
```diff
@@ -32,11 +35,17 @@ public RemoveAnnotation(Location location, String annotation) {
...
+    // We only insert annotations with their simple name, therefore, we should only remove
+    // the annotation if it matches with the simple name (otherwise, the annotation was not injected
+    // by the core module request and should not be touched).
+    AnnotationExpr annotationExpr = new MarkerAnnotationExpr(annotationSimpleName);
+    NodeList<AnnotationExpr> annotations = node.getAnnotations();
+    annotations.removeIf(annotationExpr::equals);
   }
 
   @Override
+  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
nimakarimipour: Should have avoided these unnecessary changes will avoid in future.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
```diff
@@ -32,11 +35,17 @@ public RemoveAnnotation(Location location, String annotation) {
   }
 
   @Override
-  public boolean apply(CompilationUnit tree) {
-    return this.location.apply(tree, annotation, false);
+  public void visit(NodeWithAnnotations<?> node) {
+    // We only insert annotations with their simple name, therefore, we should only remove
+    // the annotation if it matches with the simple name (otherwise, the annotation was not injected
+    // by the core module request and should not be touched).
+    AnnotationExpr annotationExpr = new MarkerAnnotationExpr(annotationSimpleName);
```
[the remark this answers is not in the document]
nimakarimipour: Updated comment in removal annotation 527f375a90797c5e7dd7d296665b5a228e75508f.

core/src/test/java/edu/ucr/cs/riple/core/AnalysisModeTest.java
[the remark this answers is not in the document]
nimakarimipour: Added a test to clarify this f441ef49d569c776988a1d217c1b3ec744fbe883. Thank you 🙂

core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: Added the comment.

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: 808eff84d09666d5dc4bb7885bcb1e0e958303b0

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: 559b012597d457467dd96aa995a4947f3faf61a3

I updated the comment, Yes the logic seems correct to me. Sure in future we can have follow ups for marking a class as `@NullUnmarked` instead of marking most of its methods.

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: I didnt fully understand what to do here, the remaining errors are mostly due their corresponding fixes get rejected (error in downstream or effect more than 0) or no corresponding fix exists.

nimakarimipour:
@lazaroclapp Thank you for the review, this is ready for another review.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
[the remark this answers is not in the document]
lazaroclapp: Technically, you needed this functionality already for `@SuppressWarnings` anyways, but happy with the `CustomNull` test. Doesn't resolve the issue of adding extra info as a comment (will write more in response to another comment thread below), but that might need to be a follow up.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddAnnotation.java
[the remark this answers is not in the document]
lazaroclapp: Um, are you sure? I don't see the suppression in the "before" code, either at the method or class level...

injector/src/main/java/edu/ucr/cs/riple/injector/changes/Change.java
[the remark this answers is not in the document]
lazaroclapp: (For _this_ class, the above is true :) )

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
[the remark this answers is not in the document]
lazaroclapp: Again, I don't think this suppression was here before, either at method or class level, unless I am missing something...

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
lazaroclapp: So, what I mean here is, since we can't actually have:

```
@NullUnmarked("Some text")
public Object foo {
```

But we want to be able to inject some sort of explanation for why we are unmarking. I think we need either:

```
// Explanation: Adding @Nullable here causes 102 errors in downstream dependencies
@NullUnmarked
public Object foo {
```

Or, even *better*:

```
// [NULLAWAY_ANNOTATOR_ISSUE_{UUID}]
@NullUnmarked
public Object foo {
```

with a file (**issues.tsv**?) on the side mapping `NULLAWAY_ANNOTATOR_ISSUE_{UUID}` to a description of the reason the auto-annotator added the unmarked annotation (the idea here being that a simple search and replace can: a) create a new JIRA task and copy in the reason from `issues.tsv`, b) search and replace the code, changing the corresponding `NULLAWAY_ANNOTATOR_ISSUE_{UUID}` string to a link to the issue.

That said, I think this can be a follow up for this diff, and very possible a post-internship task for either of us, given the timeline.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
```diff
@@ -177,10 +181,204 @@ public void onParameterCustomAnnotationTest() {
...
+            "   public void test(Object f) {",
+            "      h = f;",
+            "   }",
+            "}")
+        .addChanges(
+            new AddSingleElementAnnotation(
+                new OnMethod("Super.java", "com.edu.Super", "test(java.lang.Object)"),
+                "edu.ucr.CustomNull",
+                "arg3",
+                true))
```
lazaroclapp, line 381: Appreciate the comprehensive test suite here, btw!

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
```diff
@@ -0,0 +1,150 @@
...
+ */
+public class AddSingleElementAnnotation extends AddAnnotation {
+  /** Argument of the annotation. */
+  private final String argument;
+
+  /**
+   * If true, if an annotation with the same name exists, the existing annotation should include the
+   * passed argument, otherwise, a new annotation will be injected to the node.
+   */
+  private final boolean collapseArguments;
```
lazaroclapp, line 51: Why call it this and not just `repeatable` to follow JLS terminology? (It would imply reversing the logic, of course, if `repeatable=true` then a _new_ annotation is added, if it's false, then the expression inside is expanded into an array expression).

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
```diff
@@ -0,0 +1,150 @@
...
+
+  public AddSingleElementAnnotation(
+      Location location, String annotation, String argument, boolean collapseArguments) {
+    super(location, annotation);
+    Preconditions.checkArgument(
+        argument != null && !argument.equals(""),
+        "argument cannot be null or empty, use AddAnnotation instead.");
```
lazaroclapp, line 58: I understand not allowing null (Is this code being checked by NullAway, though? Because if so, we don't need the runtime check 😉 ). But, why is empty string not allowed? `@Foo("")` is a valid annotation if `Foo` is a Single-Element Annotation.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
```diff
@@ -0,0 +1,150 @@
...
+  public void visit(NodeWithAnnotations<?> node) {
+
+    StringLiteralExpr argumentExp = new StringLiteralExpr(argument);
+    // Check all existing annotations with arguments.
+    boolean argumentExists =
+        node.getAnnotations().stream()
+            .anyMatch(
+                annot -> {
+                  if (!(annot.getNameAsString().equals(annotation)
+                      || annot.getNameAsString().equals(annotationSimpleName))) {
```
lazaroclapp, line 73: Let's not change this in this PR, but it's a bit weird that `annotationSimpleName` (and `annotation`) are fields of `Change`, rather than of `AddAnnotation`.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
```diff
@@ -0,0 +1,150 @@
...
+          "Cannot update existing annotation with type: "
+              + existingAnnotation.getClass()
+              + ". Please file an issue with the expected behaviour at: https://github.com/nimakarimipour/NullAwayAnnotator/issues. Thank you!");
+    }
+
+    SingleMemberAnnotationExpr singleMemberAnnotationExpr =
+        (SingleMemberAnnotationExpr) existingAnnotation;
+    ArrayInitializerExpr updatedMemberValue = new ArrayInitializerExpr();
+    NodeList<Expression> nodeList = new NodeList<>();
+    nodeList.add(argumentExp);
```
lazaroclapp, line 117: Nit: I'd move this below the two ifs below, that way we are appending to argument lists, rather than prepending

lazaroclapp (commented):
Nice. I appreciate the refactoring into `AddMarkerAnnotation` and `AddSingleElementAnnotation`. Some (hopefully minor) comments below. Please note that there are also a few - even smaller - nits in the replies to comment threads from the last pass.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddAnnotation.java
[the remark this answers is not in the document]
nimakarimipour: Oh yes, sorry. This one wasn't there before, I just added it to silent the constant IDE warning.

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: Sure, I can do that as a follow up. We can discuss this before the internship ends on what is best output.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
nimakarimipour: Sure 🙂

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: Thanks for the suggestion, that makes sense. d3c427caf60b8a30695947fa7e2e6b053ef92cb1

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: Not now, but It will be checked by `NullAway` as a long-term plan mentioned in #21 in future. I think it does not make sense to add the annotation as `@Foo("")`. Maybe we can force it to use `MarkerAnnotation` in such cases ? However, if you think still it might be useful in some scenarios to support such injections, please let me know and I will remove the guard. Thank you.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: Sure, will do that.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: fbc738bbfdfb9570efe8a917b36ba7fcfc24fc91

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: #70

nimakarimipour:
@lazaroclapp Thanks for the review, this is ready for another review.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
lazaroclapp, line 49: This one should still be removed.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
lazaroclapp: Let's leave it like this for now. I don't think there is anything in the Java spec that prevents you from having a single element annotation with the empty string as its `value` field, and there is definitely a semantic difference between empty-string vs no-value, but I don't think we care for any of the foreseeable use cases of the auto-annotator, so let's leave the check in for now.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
```diff
@@ -95,13 +95,13 @@ public void visit(NodeWithAnnotations<?> node) {
       return;
     }
 
-    // Annotation with the same name exists, if collapse is off, add it directly.
-    if (!collapseArguments) {
+    // Annotation with the same name exists, if repeatable is off, add it directly.
```
lazaroclapp, line 98:
```
// Annotation with the same name exists, but the annotation is repeatable, add it directly.
```

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
```diff
@@ -95,13 +95,13 @@ public void visit(NodeWithAnnotations<?> node) {
...
+    if (repeatable) {
       addAnnotationExpressionOnNode(node, argumentExp);
       return;
     }
 
-    // Annotation with the same name exists, collapse is on, update it.
+    // Annotation with the same name exists, repeatable is on, update it.
```
lazaroclapp, line 104:
```
// Annotation with the same name exists and is not repeatable, update it.
```

lazaroclapp (commented):
Very minor nits remaining: changes to comments and one unnecessary suppression

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: Ok great sounds great. Will update it in future if needed.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
nimakarimipour: dc472d03eebba3dd33485b776a0b98e35f155266

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: Updated it.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddSingleElementAnnotation.java
nimakarimipour: Updated it. Thanks for the suggestion.

nimakarimipour:
@lazaroclapp Thank you for the review, will land it once the CI pass.

lazaroclapp (approved):
LGTM! :)

PR #74, opened 2022-09-17
Fix #73 by extending list of NullUnmarked injections.

### LINKED ISSUE #73: Force resolve cannot suppress errors for nullable param pass in field initializations.
**Describe the bug**
We currently support adding SuppressWarnings and NullUnmarked annotation to silence remaining errors through #69. After the outer loop completes, Annotator will run NullAway another time and surpasses all remaining errors with either `@NullUnmarked` or `@SuppressWarnings`. All remaining errors store their corresponding regions where that error is triggered. For all triggered regions where the enclosing method is not `"null"`, we mark them with `@NullUnmarked` annotation. Remaining errors are triggered in field declaration regions of classes. For field declarations, if they are assigned to `@Nullable` we suppress it by adding `@SupreessWarnings("NullAway")` and if they are nit initialized we add `@SupreessWarnings("NullAway.Init")`. This does not cover all cases, such as example below:

```java
class Foo {
     Bar f = new Bar(null);
}
``` 

In this case we will receive the error: `passing @Nullable where @Nonnull was expected` and the triggered regions are `class: Foo method: "null"`. This case is not covered in our current strategy. We should mark the called constructor with `@NullUnmarked`.

**To Reproduce**
Run with force resolve mode activated on the code below:
```java
class Foo {
     Bar f = new Bar(null);
}
```

**Expected behavior**
Mark `Bar(Object)` as `@NullUnmarked`.

**Stack trace**
Not-Applicable.

 - OS: x.
 - Version: x. 
 - Annotator Version:  v1.3.3-SNAPSHOT.

**Additional context**
Not-Applicable.

nimakarimipour:
This PR resolves #73 by adding `@NullUnmarked` annotations for called methods in field initializations that received a `@Nullable`.

### DIFF
--- a/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -267,7 +267,9 @@ public class Annotator {
                     // Just a sanity check.
                     return null;
                   }
-                  // For methods invoked in an initialization region, where the error is that `@Nullable` is being passed as an argument, we add a `@NullUnmarked` annotation to the called method.
+                  // For methods invoked in an initialization region, where the error is that
+                  // `@Nullable` is being passed as an argument, we add a `@NullUnmarked` annotation
+                  // to the called method.
                   if (error.messageType.equals("PASS_NULLABLE")) {
                     OnMethod calledMethod = error.nonnullTarget.toMethod();
                     return tree.findNode(calledMethod.method, calledMethod.clazz);
--- a/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -267,8 +267,7 @@ public class Annotator {
                     // Just a sanity check.
                     return null;
                   }
-                  // method called in initialization region of class and received null for
-                  // arguments.
+                  // For methods invoked in an initialization region, where the error is that `@Nullable` is being passed as an argument, we add a `@NullUnmarked` annotation to the called method.
                   if (error.messageType.equals("PASS_NULLABLE")) {
                     OnMethod calledMethod = error.nonnullTarget.toMethod();
                     return tree.findNode(calledMethod.method, calledMethod.clazz);
--- a/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -267,6 +267,8 @@ public class Annotator {
                     // Just a sanity check.
                     return null;
                   }
+                  // method called in initialization region of class and received null for
+                  // arguments.
                   if (error.messageType.equals("PASS_NULLABLE")) {
                     OnMethod calledMethod = error.nonnullTarget.toMethod();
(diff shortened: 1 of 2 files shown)


core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -256,14 +257,27 @@ private void forceResolveRemainingErrors(
...
+                    OnMethod calledMethod = error.nonnullTarget.toMethod();
+                    return tree.findNode(calledMethod.method, calledMethod.clazz);
+                  }
+                  return null;
+                })
             .filter(Objects::nonNull)
             .map(node -> new AddMarkerAnnotation(node.location, config.nullUnMarkedAnnotation))
             .collect(Collectors.toSet());
+    // Collect methods received a nullable i their nonnull parameters where called in field
+    // declaration regions of classes (method == "null).
```
msridhar, line 280: Is this the right place for this comment?  I don't understand how it relates to the code

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -299,6 +313,7 @@ private void forceResolveRemainingErrors(
             // Exclude already annotated fields with a general NullAway suppress warning.
             .filter(f -> !suppressWarningsAnnotations.contains(f))
             .collect(Collectors.toSet());
+
```
msridhar, line 316: extraneous blank line

msridhar (commented):
Mostly LGTM, just a couple comments

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: 953786d6bb704e12d25d60a1ac190f2c3a164737 sorry for that. Removed the comment. 953786d6bb704e12d25d60a1ac190f2c3a164737

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: 67a4dbf6b6f9b318f4782f405f1966de2f7666a3

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -256,11 +257,24 @@ private void forceResolveRemainingErrors(
...
+                error -> {
+                  if (!error.encMethod().equals("null")) {
+                    return tree.findNode(error.encMethod(), error.encClass());
+                  }
+                  if (error.nonnullTarget == null) {
+                    // Just a sanity check.
+                    return null;
+                  }
+                  // method called in initialization region of class and received null for
+                  // arguments.
```
msridhar, line 271: Suggested wording change, for clarity:
```suggestion
                  // For methods invoked in an initialization region, where the error is that `@Nullable` is being passed as an argument, we add a `@NullUnmarked` annotation to the called method.
```

msridhar (approved):
Approved with one comment

nimakarimipour:
@msridhar Thank you for the review, will land it now.

PR #83, opened 2022-09-23
Add missing equals to changes

nimakarimipour:
This PR adds refines `equals()` definition for `Change` subclasses.


nimakarimipour:
@msridhar Thank you very much for that. Will add it now.

PR #94, opened 2022-11-12
Fix URIs in MethodDeclarationTree nodes

### LINKED ISSUE #93: Wrong uri in MethodDeclarationTree nodes
**Describe the bug**
For `TypeAnnotatorScanner`

`URI`s are fetched from the `VisitorState` instances. For sub methods, the `URI` is fetched from a state while it is pointing at the super method and a wrong `URI` is fetched.

**To Reproduce**
Run `TypeAnnotatorScanner` on modules with base and super class.

**Expected behavior**
All nodes in the tree have the uri to the containing source file.

**Stack trace**
Not-Applicable

 - Annotator Version [e.g. v1.3.4-SNAPSHOT # master]

**Additional context**
This will also prevent Annotator to correctly compute annotations for `@NullUnmarked` injections.

nimakarimipour:
This PR resolves #93 by passing the relevant `VisitorState` for fetching the `URI`.

#### Note
This will enable to Annotator to correctly add `@NullUnmarked` annotations as generating annotation requests requires getting URIs from `MethodDeclarationTree`.

### DIFF
--- a/type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/TypeAnnotatorScanner.java
+++ b/type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/TypeAnnotatorScanner.java
@@ -106,9 +106,10 @@ public class TypeAnnotatorScanner extends BugChecker
       return Description.NO_MATCH;
     }
     Symbol.MethodSymbol methodSymbol = ASTHelpers.getSymbol(tree);
-    MethodInfo methodInfo = MethodInfo.findOrCreate(methodSymbol, state, context);
+    MethodInfo methodInfo = MethodInfo.findOrCreate(methodSymbol, context);
     methodInfo.findParent(state, context);
     methodInfo.setAnnotation(config);
+    methodInfo.setURI(state);
     List<Boolean> paramAnnotations = new ArrayList<>();
     for (int i = 0; i < methodSymbol.getParameters().size(); i++) {
       paramAnnotations.add(SymbolUtil.paramHasNullableAnnotation(methodSymbol, i, config));
--- a/type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
+++ b/type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodInfo.java
@@ -26,6 +26,7 @@ package edu.ucr.cs.riple.scanner.out;
 
 import com.google.common.base.Preconditions;
 import com.google.errorprone.VisitorState;
+import com.sun.source.tree.CompilationUnitTree;
 import com.sun.tools.javac.code.Symbol;
 import edu.ucr.cs.riple.scanner.Config;
 import edu.ucr.cs.riple.scanner.ScannerContext;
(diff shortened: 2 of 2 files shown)


type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/TypeAnnotatorScanner.java
```diff
@@ -106,9 +106,10 @@ public Description matchMethod(MethodTree tree, VisitorState state) {
       return Description.NO_MATCH;
     }
     Symbol.MethodSymbol methodSymbol = ASTHelpers.getSymbol(tree);
-    MethodInfo methodInfo = MethodInfo.findOrCreate(methodSymbol, state, context);
+    MethodInfo methodInfo = MethodInfo.findOrCreate(methodSymbol, context);
     methodInfo.findParent(state, context);
     methodInfo.setAnnotation(config);
+    methodInfo.setURI(state);
```
lazaroclapp, line 112: So, I understand why we are adding the check for a `null` `tree.getSourceFile()` (generated classes and so on, I assume), but why are we moving this computation from the constructor to this `setURI(...)` method? Do we expect his to be called multiple times? Or a long time after first calling `MethodInfo.findOrCreate(...)`?

It seems more natural to me that `MethodInfo.uri` is `final` and set by the constructor.

Either way, that's partly a matter of style, so I am not opposed to either way.

lazaroclapp (commented):
This PR notes it is fixing something, but I am not sure what the issue was previous to the fix. A `null` dereference of `CompilationUnitTree.getSourceFile()`? Is it reasonable to have a test case that wouldn't have passed before but does now?

type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/TypeAnnotatorScanner.java
nimakarimipour: Yes, the problem is that the actual call to `setURI` is called after the initial object creation via `findOrCreate`. That happens when a sub-method is visited sooner than the the super-method and this is the reason we were having wrong `URI`s in our nodes.

nimakarimipour: Please note, that while visiting a `sub-method` a super method node is `created / reserved` and then we fill other info for that node when we are visiting it via `match` methods.

nimakarimipour: @lazaroclapp I checked adding a unit test with mocking for this case. It is not straightforward adding this test without changing the constructors (Requires changing the return result of `CompilationUnitTree` passed from Error Prone). Considering that the objective of this PR is not handling `tree.getSourceFile()` returning `null` but passing the correct `VisitorState` to method node (not the `VIsitorState` pointing to super-method). Would you please tell me if I should still try to make a unit test for this ? Thank you.

lazaroclapp: Ah, I think I understand this better now. `findOrCreate(...)` is called both here and also from `findParent(...)`, but when called from `findParent(...)` the `VisitorState` will correspond to the overridding method, while the `methodSymbol` will be the method being overridden. Moving the logic of `setURI(...)` to a separate call that only happens when directly visiting the given method (via `matchMethod`), ensures that the `VisitorState` is looking at the file where `methodSymbol` is declared. Correct?

lazaroclapp (approved):
LGTM.

type-annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/TypeAnnotatorScanner.java
nimakarimipour: @lazaroclapp Yes exactly. This is the issue that has been resolved.

nimakarimipour:
> LGTM.

Great, thank you for the review. Will land it now.

PR #95, opened 2022-11-12
Add suppressWarnings on fields

nimakarimipour:
~~This PR is built upon #91 and #74~~ (#91 and #74 is landed)

#### CI
~~Please note that CI will fail for this PR as it requires release of `NullAway:0.10.5`.~~
`NullAway:0.10.5` is released and CI should pass.

-----

This PR adds `@SuppressWarnings("NullAway")` on fields that cannot get annotated as `@Nullable` due to their impacts and a `@Nullable` value is used as the initializer.


nimakarimipour:
@msridhar @lazaroclapp FYI, This PR is rebased and ready for review. Thank you 🙂.

core/src/test/resources/fielderrorregion/Foo.java
```diff
@@ -0,0 +1,29 @@
...
+  Bar f1;
+  Bar f2, f3;
+  @Nullable Bar nullableBar;
+  Bar f4 = nullableBar.process(new Object());
+
+  Foo() {
+    this.f2 = new Bar();
+  }
+
+  public void run1() {
```
msridhar, line 16: Why do we need both `run1` and `run2`?

msridhar (approved):
There seem to be some minor unrelated changes here but I am still fine with landing.

core/src/test/resources/fielderrorregion/Foo.java
nimakarimipour: Yes exactly that was needed to produce more errors to make sure `@Nullable` is not inferred.

Regarding `f4`, I am not sure if I fully understood the question, making `f4` will resolve one error which is assigning `@Nullable` and triggers 2 new errors which is deref of `f4` therefore the actual effect should be `1`. I checked it manually the effect is 1 and correct. Please let me know if I am missing something.

lazaroclapp: Ok, then, but, by that logic: shouldn't it be the same number for `f1` and `f3`? There is also one error being resolved (the missing initialization error) and two being added. But the corresponding `TReport`s have `2` as the effect. Either way, I don't understand the discrepancy between the number for `f4` and the other two.

core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -297,6 +303,65 @@ public void deactivateInferenceTest() {
...
+                    "test.Bar",
+                    "process(java.lang.Object)"),
+                "org.jspecify.nullness.NullUnmarked"),
+            new AddSingleElementAnnotation(
+                new OnField(srcRoot.resolve("Foo.java").toString(), "test.Foo", Set.of("f4")),
+                "SuppressWarnings",
+                "NullAway",
+                false),
+            new AddSingleElementAnnotation(
+                new OnField(srcRoot.resolve("Foo.java").toString(), "test.Foo", Set.of("f0")),
```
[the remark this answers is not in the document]
nimakarimipour: Thank you very much for catching this bug. It was both annotated as `@Nullable` and `@SuppressWarnings`. The `@SuppressWarnings` was added because of the passing `null` error but that error is resolved with `@NullUnmarked` injection but was selected again later in `SuppressWarnings` selection logic. This is resolved now and `f0` is just annotated as `@Nullable`.

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -267,42 +267,49 @@ private void forceResolveRemainingErrors(
...
-                  }
                   // For methods invoked in an initialization region, where the error is that
                   // `@Nullable` is being passed as an argument, we add a `@NullUnmarked` annotation
                   // to the called method.
-                  if (error.messageType.equals("PASS_NULLABLE")) {
-                    OnMethod calledMethod = error.nonnullTarget.toMethod();
+                  if (error.messageType.equals("PASS_NULLABLE")
+                      && error.nonnullTarget != null
+                      && error.nonnullTarget.isOnParameter()) {
+                    OnParameter calledMethod = error.nonnullTarget.toParameter();
```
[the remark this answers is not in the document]
nimakarimipour: Thank you for noticing this 3301cdacf6cf33a7bf6952fd7cd27346678bc128. It is marking a method `@NullUnmarked` that received a `@Nullbale` for parameter which annotater rejected its parameter to be `@Nullable`. Renamed it.

core/src/test/resources/fielderrorregion/Foo.java
nimakarimipour: Actually the difference is that making method `f1` or `f3` individually will not resolve the initialization error. The error still remains after the injection. It needs both annotation to be inserted to resolve the error. It only creates two new errors and does not resolve any error, hence the effect is `+2`.

nimakarimipour:
@msridhar @lazaroclapp Thank you for your comments and review. This is ready for another round. Thank you.

core/src/test/resources/fielderrorregion/Foo.java
lazaroclapp: What is the initialization error in question here?

Is it one error for `f1` not being initialized and one error for `f3` not being initialized, or is the generic error at `Foo()` saying some fields are not initialized by the constructor?

I thought we were serializing individual initialization errors. If we aren't, then wouldn't any class with 2 or more constructors and at least one field which is a) not initialized, b) treated as non-null in more than one place, then result in all non-initialized fields being marked `@SuppressWarnings(...)`? Even when many could otherwise be marked `@Nullable`? It seems like the auto-annotator should be able to consider each uninitialized field as a separate error...

To test this, could we have an example analogous to `f1` but without the explicit dereferences and make sure that is actually determined to be `@Nullable`?

Or am I missing a different reason why adding `@Nullable` to `f1` or `f3` doesn't remove the initialization error for that field?

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -294,6 +294,10 @@ private void forceResolveRemainingErrors(
                   if (!error.getRegion().isOnField()) {
                     return false;
                   }
+                  if (error.messageType.equals("PASS_NULLABLE") && error.getRegion().isOnField()) {
```
lazaroclapp, line 297: Isn't the second condition already guaranteed by the check right above? (i.e. at this point `error.getRegion().isOnField()` is known to be `true` or we would have hit the early return on line 295)

lazaroclapp (commented):
The last couple of changes make sense, but I still have some questions about how we are counting the effect / errors resolved, when dealing with field initialization...

core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: Yes, that makes sense, removed it. Thank you c3b2464c2006a93255f580fa94b926d7ae60f31c

core/src/test/resources/fielderrorregion/Foo.java
nimakarimipour: @lazaroclapp Actually we only serialize one error for initialization errors per constructor which can consist of multiple fields not being initialized. The main purpose of Annotator is to minimize the number of errors the user sees from NullAway output, therefore, we do not split an error coming from NullAway into multiple errors for each uninitialized field. However, the scenario which is mentioned does not happen as we serialize a fix object for each uninitialized field and nullability of each one is investigated individually. I added a unit test for that, please see 8fb982c6cb48a22c1958c6ae6cd2d6d86cafbc68

In this unit test, 

- `f1`, `f2` and `f3` are not initialized by constructors and are candidates for `@Nullable`. 
-  Marking `f1`, `f2` or `f3` as `@Nullable` individually will not resolve the one initialization error complaining about the other two fields.
- `f3` has both conditions, not initialized and treated as `@Nonnull` in more than one place.
- `f1` and `f2` are determined to be `@Nullable`.
- `f3` is determined to be `@Nonnull` and instead is annotated with `SuppressWarnings("NullAway.Init")`
- `f4` was not involved in initialization errors (just `assign nullable` error) and cannot be `@Nullable`, therefore it is marked with `@SuppressWarnings("NullAway")`.

core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -357,6 +349,58 @@ public void errorInFieldDeclarationForceResolveTest() {
...
+                "NullAway.Init",
+                false),
+            new AddSingleElementAnnotation(
+                new OnField(pathToFoo, "test.Foo", Set.of("f4")),
+                "SuppressWarnings",
+                "NullAway",
+                false));
+    Assert.assertEquals(
+        expectedAnnotations, Set.copyOf(coreTestHelper.getConfig().log.getInjectedAnnotations()));
+  }
```
lazaroclapp, line 402: 👍

core/src/test/resources/fielderrorregion/Foo.java
```diff
@@ -0,0 +1,29 @@
+package test;
+
+import javax.annotation.Nullable;
+
+public class Foo {
+  Bar f0 = new Bar().process(null);
+  Bar f1;
+  Bar f2, f3;
+  @Nullable Bar nullableBar;
+  Bar f4 = nullableBar.process(new Object());
```
lazaroclapp, line 10: New test case looks great to me! Could we still add `Bar f5;` without initialization or dereference to this single constructor test case and make sure it's marked as `@Nullable`. Just for the sake of testing all cases. If I understood the explanation correctly, then I fully expect it to work and then we can land this 🙂

lazaroclapp (commented):
One small test case request and then this looks ready to me!

core/src/test/resources/fielderrorregion/Foo.java
nimakarimipour: Sure, please find it 970a2abe3aa4ff841faaf515e59c00c68d3ad602

lazaroclapp (approved):
Looks good to land if/when CI passes! 🚀

PR #96, opened 2022-11-14
Release scripts

nimakarimipour:
This PR adds release configurations for `core` and `type-annotator-scanner` modules.

Note: `core` will be renamed to `annotator-core` as a followup.

### DIFF
--- a/type-annotator-scanner/build.gradle
+++ b/type-annotator-scanner/build.gradle
@@ -22,6 +22,8 @@
  * THE SOFTWARE.
  */
 
+import com.vanniktech.maven.publish.SonatypeHost
+
 buildscript {
     repositories {
         mavenCentral()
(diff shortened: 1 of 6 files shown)


type-annotator-scanner/build.gradle
```diff
@@ -36,14 +38,21 @@ buildscript {
 plugins {
     id "net.ltgt.errorprone" version "2.0.1" apply false
     id 'com.github.sherter.google-java-format' version '0.9'
+    id 'maven-publish'
```
msridhar, line 41: Minor: I don't think we need to add `maven-publish` explicitly if we are already adding `com.vanniktech.maven.publish`?  I think including the latter should include the former automatically.  But I know this is brittle, so if this is working we can leave it this way.

msridhar (approved):
LGTM

type-annotator-scanner/build.gradle
nimakarimipour: Yes this is working for now, Sure I can test it in follow up cycles.

PR #101, opened 2022-11-30
Refactor core logic (preparation for modeling implementation)

nimakarimipour:
This PR refactors the logic implemented in `annotator-core` to prepare the infrastructure for implementing the new modeling approach. 

Please find the list of changes below:
1. `Explorer` is renamed to `Evaluator` which is responsible for creating a set of `reports` for the given `fixes`.
2. Logic implemented inside `Explorer` for `injecting` / `removing` fixes and rerunning compiler is extracted to a separate classes with interface called `ConflictGraphProcessor`.
3. `Evaluator` classes uses `ConflictGraphProcessor` instances to compute the effect of fixes.
4. `OptimizedExplorer` is removed completely.
5. `ExhaustiveExplorer` is renamed to `VoidEvaluator`.
6. `ConflictGraphProcessor` has two subclasses: `SequentialConflictGraphProcessor` and `ParallelConflictGraphProcessor` which uses graph coloring techniques.
7. `BasicEvaluator` is subclass of `Evaluator` which injects a fix tree completely to capture impacts.
8. `ModeledEvaluator` (which will be added in upcoming PR) is subclass of `Evaluator` that injects only undiscovered fixes and uses cached results to capture impacts.

Below is the settings for desired experiments:
1. Modeling enabled Graph coloring enabled: `ModeledEvaluator` plugged with `ParallelConflictGraphProcessor`.
2. Modeling enabled Graph coloring disabled: `ModeledEvaluator` plugged with `SequentialConflictGraphProcessor `.
3. Modeling disabled Graph coloring enabled: `BasicEvaluator` plugged with `ParallelConflictGraphProcessor`.
4. Modeling disabled Graph coloring disabled:  `BasicEvaluator` plugged with `SequentialConflictGraphProcessor `.


annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/AbstractEvaluator.java
```diff
@@ -0,0 +1,87 @@
...
+public abstract class AbstractEvaluator implements Evaluator {
+
+  /** Annotator config. */
+  protected final Config config;
+  /** Conflict graph to storing unprocessed fixes. */
+  protected final ConflictGraph graph;
+  /** Depth of analysis. */
+  protected final int depth;
+  /** Graph processor to process the graph. */
+  protected ConflictGraphProcessor analyzer;
```
lazaroclapp, line 45: Why is this field named `analyzer` and not e.g. `processor` or even `graphProcessor`? Wouldn't that be clearer?

lazaroclapp (approved):
So, there is a bound here in my understanding of both the original and new architecture for the auto-annotator, but both the list of refactoring and the individual code changes make sense to me. Given all tests pass, this refactoring LGTM! 🙂 

Just to be sure, are there any configurations/combination of CLI flags that worked on 1.3.4 which will _not_ work after this refactoring? Fix-at-a-time? Any others? (If so, we might want to make sure that we fail early if we detect any config we no longer support, rather than silently ignore flags that used to do something)

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/AbstractEvaluator.java
nimakarimipour: That should have been `processor`. Thank you for catching this. c92be6aa18314d9eabca233a27bd3d1bb70ee5ae

nimakarimipour:
> So, there is a bound here in my understanding of both the original and new architecture for the auto-annotator, but both the list of refactoring and the individual code changes make sense to me. Given all tests pass, this refactoring LGTM! 🙂
> 
> Just to be sure, are there any configurations/combination of CLI flags that worked on 1.3.4 which will _not_ work after this refactoring? Fix-at-a-time? Any others? (If so, we might want to make sure that we fail early if we detect any config we no longer support, rather than silently ignore flags that used to do something)

@lazaroclapp Thank you for the review, If there is anything you need to discuss further please let me know. Also to answer your last question, __all__ supporting features in `1.3.4` will be still supported in the new architecture as well. 🙂

PR #105, opened 2022-12-03
Add resolving fixes to error instances.

nimakarimipour:
This PR updates `Error` data structs to store set of fixes which can resolve the reporting error instance. This change prepares the infra for the new modeling approach.

Previously location of the element involved in a pseudo-assignment was stored for each reporting error if such elements existed. With this PR, a list of `fixes` that should be applied to resolve the error is stored.


nimakarimipour:
Closing this PR as is it is re-implemented in #112

PR #107, opened 2022-12-07
Add -Werror for javac

nimakarimipour:
#### Low Priority

This PR adds `-Werror` javac flag to builds and gets rid of all current warnings.

@msridhar @lazaroclapp this is low priority but just wanted to get rid of warnings 🙂.

### DIFF
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/ConflictGraph.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/ConflictGraph.java
@@ -74,7 +74,7 @@ public class ConflictGraph {
    * in the same group. A greedy algorithm is used to find the solution.
    */
   // TODO: Remove SuppressWarnings below later.
-  @SuppressWarnings("JdkObsolete")
+  @SuppressWarnings({"JdkObsolete", "unchecked"})
   public void findGroups() {
     this.groups.clear();
     Collection<Node> allNodes = nodes.values();
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,8 @@ subprojects { proj ->
     proj.dependencies {
         errorprone deps.build.errorProneCore
         errorproneJavac deps.build.errorProneJavac
+        // just to pass wierd javac builds.
+        compileOnly 'com.google.errorprone:error_prone_core:2.4.0'
         if(proj.name != "ban-mutable-static"){
             annotationProcessor project(":checks:ban-mutable-static")
         }
@@ -66,6 +68,7 @@ subprojects { proj ->
                 }
             }
         }
+        options.compilerArgs += '-Werror'
     }
 
     repositories {
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ subprojects { proj ->
     proj.dependencies {
         errorprone deps.build.errorProneCore
         errorproneJavac deps.build.errorProneJavac
-        // just to pass wierd javac builds.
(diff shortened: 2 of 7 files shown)


annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/ConflictGraph.java
```diff
@@ -74,7 +74,7 @@ public Node addNodeToVertices(Fix fix) {
    * in the same group. A greedy algorithm is used to find the solution.
    */
   // TODO: Remove SuppressWarnings below later.
-  @SuppressWarnings("JdkObsolete")
+  @SuppressWarnings({"JdkObsolete", "unchecked"})
```
msridhar, line 77: Why not just fix the error instead of suppressing?

build.gradle
```diff
@@ -48,6 +48,9 @@ subprojects { proj ->
     proj.dependencies {
         errorprone deps.build.errorProneCore
         errorproneJavac deps.build.errorProneJavac
+        // just to prevent javac warnings, we currently pass "-Werror" as compiler arguments,
+        // all warnings must be resolved.
+        compileOnly deps.build.errorProneCoreOld
```
msridhar, line 53: I don't understand this change

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
```diff
@@ -64,6 +64,7 @@ public Modification visit(NodeWithAnnotations<?> node, Range range) {
   }
 
   @Override
+  @SuppressWarnings("unchecked")
```
msridhar, line 67: Again why suppress

msridhar (commented):
couple questions

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/ConflictGraph.java
nimakarimipour: I didn't know how I can resolve the warning for creating array of `LinkedLists`. Since I was confident that there will be no problem with the code I just silenced it. 

However, I just changed it to use List of Lists 6517fba69335d5a5e7038d61722e0065a5eec73a, I don't think this will impact the performance. But if you think this will actually impact the performance, please let me know and I will undo this. Thank you.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
nimakarimipour: @msridhar The only way to resolve this warning is to change the json dependency to `org.json` from `com.googlecode.json-simple:json-simple` since it is compiled with a very old javac (the latest release is for 2012). I am definitely interested in updating this dependency to use `org.json` but that will include a lot of changes which makes this PR pretty large. Should I do it in a follow up PR with an issue for now to track it ?

nimakarimipour:

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveAnnotation.java
nimakarimipour: Just created a followup PR that removes the added `@SuppressWarnings` in this PR and all existing ones by changing dependency to `org.json`. #123.

msridhar:
@nimakarimipour do you still need a review on this one?

PR #109, opened 2022-12-09
Add annotations without indentation

nimakarimipour:
This PR updates injector to add annotations without indentation considerations since it adds unnecessary complications. All indentations can fixed by the using formatter in the target project in the end. It will also simplify #103.


injector/src/test/java/edu/ucr/cs/riple/injector/RemovalTest.java
```diff
@@ -162,7 +162,7 @@ public void removeMethodInlineAnnot() {
             "package edu.ucr;",
             "import edu.custom.Annot;",
             "public class Test {",
-            "   public@Annot Object test() {",
+            "   public @Annot Object test() {",
```
msridhar, line 165: Is this change because we changed how we insert annotations to have a leading space?

injector/src/test/java/edu/ucr/cs/riple/injector/RemovalTest.java
```diff
@@ -187,7 +187,7 @@ public void removeParamAnnotEndingWithUnderlineExists() {
             "import javax.annotation.Nullable;",
             "import com.customAnnot.AnnotationWithNonAlphanumericSuffix$$$$_;",
             "public class Test {",
-            "   public void foo(@AnnotationWithNonAlphanumericSuffix$$$$_@Nullable Object o) { }",
+            "   public void foo(@AnnotationWithNonAlphanumericSuffix$$$$_ @Nullable Object o) { }",
```
msridhar, line 190: same as previous question

injector/src/test/resources/simple.java
```diff
@@ -1,5 +1,4 @@
 package injector;
-
```
msridhar, line 2: why this change?

injector/src/test/resources/simple_expected_1.java
```diff
@@ -1,12 +1,10 @@
 package injector;
 import javax.annotation.Nullable;
-
```
msridhar, line 3: why this change?

msridhar (approved):
Looks great to me!  I love a change that deletes so much code :-)  Just a couple of minor questions but I'm also ok with landing

injector/src/test/java/edu/ucr/cs/riple/injector/RemovalTest.java
nimakarimipour: Yes exactly. With new implementation we reduced scope of annotations Injector can remove since we only remove annotations the injector added itself. And the `@Annot` in this code is not how Injector adds an annotation.

injector/src/test/java/edu/ucr/cs/riple/injector/RemovalTest.java
nimakarimipour: Yes exactly.

injector/src/test/resources/simple.java
nimakarimipour: Left over from another change, thank you for noticing this. We used to add an extra new line for each import and this is fixed in previous PR. Brought it back. e48eb1a283e8df35293e92bc19f7e19d1b5fc038

injector/src/test/resources/simple_expected_1.java
nimakarimipour: Same as above.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
```diff
@@ -51,8 +51,7 @@ public void onFieldSuppressWarningTest() {
         .expectOutput(
             "package com.edu;",
             "public class Super {",
-            "   @SuppressWarnings(\"NullAway.Init\")",
-            "   Object h = new Object();",
+            "   @SuppressWarnings(\"NullAway.Init\") Object h = new Object();",
```
lazaroclapp, line 54: Mmm... I wonder if GJF will re-insert the new line here or not. 

Feels like we'd hope it does, since it matches how developers insert these suppressions better. Not sure if anyone minds these being in the same line, though, so I'd say this is fine for now. In the future, we might have to revert this decision and do `\n` in some cases and ` ` in others (should be one character either way so it shouldn't be too complicated to track, right?)

injector/src/test/java/edu/ucr/cs/riple/injector/ClassSearchTest.java
```diff
@@ -56,8 +56,7 @@ public void annotation_declaration() {
             "package com.test;",
             "import javax.annotation.Nullable;",
             "public @interface Main{",
-            "   @Nullable",
-            "   public String foo();",
+            "   @Nullable public String foo();",
```
lazaroclapp, line 59: Same here, most developers would expect this to be on a line above the method declaration.

injector/src/test/java/edu/ucr/cs/riple/injector/ComprehensiveTest.java
```diff
@@ -68,27 +68,21 @@ public void addFieldMethodParamTest() {
             "import javax.annotation.Nullable;",
             "public class Test {",
             "   Field foo1;",
-            "   @Nullable",
-            "   Field foo2;",
+            "   @Nullable Field foo2;",
```
lazaroclapp, line 71: OTOH, fields are probably better annotated in the same line unless they already have a large collection of annotations...

lazaroclapp (approved):
Should be fine to land, I think. My only question here is if this affects the style of the generated code _after_ GJF runs. 

I think there are some considerations long-term where we might care were the `@Nullable` annotation shows up (e.g. on it's own line above methods, but in the same line at the start for fields, unless they already have a series of annotations in which case we want to follow existing style...). Meanwhile suppressions always go in their own line. This can be something we re-consider after seeing a few internal enrollment diffs, though, because we might not care enough to make our life harder here. Also, it _might_ already be handled by GJF, but I don't think it is in all cases, since I see fields with both inline and above-line annotations in our codebase.

injector/src/test/java/edu/ucr/cs/riple/injector/AnnotationWithArgumentTest.java
nimakarimipour: Yes exactly, we can change this in future if this becomes an issue.

PR #110, opened 2022-12-14
Remove / clean up index store.

nimakarimipour:
#### This PR is build upon #113 

Following #113 set of triggered fixes is not computed per region and class `Bank<T extends Enclosed>` is only used to index errors. This PR performs a clean up and a simplification on this class.

Changes in this PR

- Rename `Bank` to `ErrorStore` which will be used to store, index and retrieve triggered errors per region.
- Remove `Enclosed` class as it only has one subclass now.


msridhar (approved):
This one LGTM

lazaroclapp (approved):
LGTM as well!

PR #106, opened 2022-12-07
Add modeling

nimakarimipour:
This PR is WIP.

PR #111, opened 2022-12-14
Add cached evaluators

nimakarimipour:
#### ~~This PR is built upon #103, #112, #113 and #110~~

This PR adds `CachedEvaluator` which uses caching techniques in computing impacts of fixes on module.

Caching is activated by default, to disable it, following flag must be passed to Annotator:
```
-dfc, --disable-fix-cache
```
It can also be controlled in JSON format for config:
```json
{
    "CACHE_IMPACT_ACTIVATION" : true
}
```

Changes in this PR:
- Adds `Impact` class which stores the set of triggered errors and set of triggered fixes from downstream for a fix.
- Adds `ImpactCache` which is Interface for a caches saving impacts of fixes on modules.
- Renames `GlobalAnalyzer` to `DownstreamImpactCache` for a better unified structure.
- Renames flag `-do, --disable-optimization` to `-dpp, --disable-parallel-processing` as we have two separate optimizations.


annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/BaseCache.java
```diff
@@ -0,0 +1,103 @@
...
+import edu.ucr.cs.riple.core.metadata.index.Fix;
+import edu.ucr.cs.riple.core.metadata.method.MethodDeclarationTree;
+import edu.ucr.cs.riple.injector.location.Location;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Base class for models.
```
msridhar, line 39:
```suggestion
 * Base class for impact caches.
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/BaseCache.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param <T> type of impacts saved in this model.
+ * @param <S> type of the map used to store impacts.
+ */
+public abstract class BaseCache<T extends Impact, S extends Map<Location, T>>
+    implements ImpactCache<T> {
+
+  /** Container holding caches entries. */
```
msridhar, line 47:
```suggestion
  /** Container holding cache entries. */
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/CachedEvaluator.java
```diff
@@ -0,0 +1,97 @@
...
+import edu.ucr.cs.riple.core.cache.Impact;
+import edu.ucr.cs.riple.core.cache.TargetModuleCache;
+import edu.ucr.cs.riple.core.evaluators.suppliers.Supplier;
+import edu.ucr.cs.riple.core.metadata.index.Error;
+import edu.ucr.cs.riple.core.metadata.index.Fix;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/** This evaluator uses modeling techniques to compute effect of applying a fix tree. */
```
msridhar, line 42:
```suggestion
/** This evaluator uses caching techniques to optimize computing the effect of applying a fix tree. */
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/CachedEvaluator.java
```diff
@@ -0,0 +1,97 @@
...
+public class CachedEvaluator extends AbstractEvaluator {
+
+  /** Model to retrieve impacts. */
```
msridhar, line 45:
```suggestion
  /** Cache to retrieve impacts. */
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/CachedEvaluator.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public CachedEvaluator(Supplier supplier) {
+    super(supplier);
+    this.cache = new TargetModuleCache(supplier.getConfig(), supplier.getMethodDeclarationTree());
+  }
+
+  @Override
+  protected void initializeFixGraph(ImmutableSet<Report> reports) {
+    super.initializeFixGraph(reports);
+    // add only fixes that are not stored in model.
```
msridhar, line 56:
```suggestion
    // add only fixes that are not stored in cache.
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Fix.java
```diff
@@ -54,10 +54,13 @@ public class Fix {
    */
   public boolean fixSourceIsInTarget;
 
+  public int count;
```
msridhar, line 57: What is this field for?  Javadoc?  I'd prefer accessor methods to a public, non-final field (and maybe `fixSourceIsInTarget` should be final?)

msridhar (commented):
I don't have the expertise to deeply review this code.  I feel like there is some new logic somewhere that deserves some comments, but I'm not sure where it is exactly :-)  These are mostly code cleanness comments.  Also, why are there no new tests with this PR?

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/CachedEvaluator.java
lazaroclapp: In either version, this is a very self-referential javadoc for what is a very tricky class. You are basically saying "a CachedEvaluator evaluates the effect of fixes using a cache" (or modeling, which doesn't mean anything specific either). But it's the interaction between this class and `TargetModuleCache` that gives you the core optimization implemented in this PR, no? So they _both_ need better documentation.

As I understand it, the core difference between this and the existing `BasicEvaluator` is that we are only adding fixes to the conflict graph if `cache::isUnknown` returns `true` for that fix. It is then the presence of elements in that conflict graph which triggers any re-runs of the underlying NullAway compilation. Correct?

Then, question: when does `cache::isUnknown` happen? And I thought even without `CachedEvaluator` we were avoiding retrying the same fixes if we already knew their impact. Is that wrong? What kind of results are we caching now that we weren't before?

 I think it would be worth adding a more detailed javadoc not only here, but also on `CachedEvaluator::initializeFixGraph` and `collectGraphResults`. Usually, an `@Override` method doesn't need its own Javadoc if it's obvious how it's implementing its superclass' method, but here we are very significantly changing what gets computed by these methods. No?

In general, there _are_ cases where we the javadoc comments are going to be short and provide little information over the method name (e.g. `getIterator()`: "get the iterator for this collection" is fine), but if you find yourself writing quick formulaic javadocs for complicated code that only mean something if you already have the full algorithm/architecture in your head, then that's usually a sign that better documentation is needed.

lazaroclapp (commented):
I think, overall, this needs clearer documentation. The architecture is also growing a series of abstractions that aren't exactly as separate as they should be and make it hard to understand the differences between the competing evaluators/caches/etc in terms of what builds actually get executed and when.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -233,11 +235,18 @@ public Config(String[] args) {
...
-    disableOptimizationOption.setRequired(false);
-    options.addOption(disableOptimizationOption);
+    // Parallel Processing
+    Option disableParallelProcessingOption =
+        new Option(
+            "dpp", "disable-parallel-processing", false, "Disables parallel processing of fixes");
+    disableParallelProcessingOption.setRequired(false);
+    options.addOption(disableParallelProcessingOption);
+
+    // Parallel Processing
```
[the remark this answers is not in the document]
nimakarimipour: 1e71a97e5605236be46c0d9d07dbd00ac3294789

annotator-core/src/main/java/edu/ucr/cs/riple/core/Report.java
```diff
@@ -268,4 +270,13 @@ public boolean isInProgress(Config config) {
     }
     return !config.bailout || localEffect > 0;
   }
+
+  public Set<Fix> getFixesForNextIteration() {
```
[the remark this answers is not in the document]
nimakarimipour: 1ba0d699695d021a11e28b6096ed957fbed51d4f

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Fix.java
nimakarimipour: Sorry, That field is a left over from previous design in PR #106 which is closed.  Removed it entirely.

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/CachedEvaluator.java
nimakarimipour: Thank you for all the notes and all the suggestions, I realized how incomplete the existing javadocs are! (mistakenly assumed the algorithm should only be discussed in the paper) I updated javadoc 0e568cb593d854a69822e638aaeb0623918a6357 and ad0da81b8de51c8bd0f33f3667576047308aef70.

The main difference between `BasicEvaluator` and `CachedEvaluator` is that `BasicEvaluator` includes the entire tree (with no exclusion) in the conflict graphs node and it does not store or use any impacts from previous iterations. `CachedEvaluator` only processes nodes that are not stored in the cache and assigns a single fix to each node.

nimakarimipour: @lazaroclapp Please let me know if I should add anything else.

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/BaseCache.java
```diff
@@ -0,0 +1,103 @@
...
+        .flatMap(impact -> impact.getTriggeredFixesOnDownstream().stream())
+        // filter fixes that are already inside tree.
+        .filter(fix -> !fixTree.contains(fix))
+        .collect(ImmutableSet.toImmutableSet());
+  }
+
+  @Override
+  public ImmutableSet<Error> getTriggeredErrors(Fix fix) {
+    T impact = fetchImpact(fix);
+    return impact == null ? ImmutableSet.of() : ImmutableSet.copyOf(impact.getTriggeredErrors());
```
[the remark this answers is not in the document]
nimakarimipour: 1be5241cedbee8803c42d4d12fa8391c810def61

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpactCacheImpl.java
```diff
@@ -22,54 +22,59 @@
...
-  /** Public APIs in the target modules that have a non-primitive return value. */
-  private final ImmutableMultimap<Integer, MethodImpact> methods;
-  /** Annotator Config. */
-  private final Config config;
-  /** Method declaration tree instance. */
-  private final MethodDeclarationTree tree;
 
-  public GlobalAnalyzerImpl(Config config, MethodDeclarationTree tree) {
-    this.config = config;
+  public DownstreamImpactCacheImpl(Config config, MethodDeclarationTree tree) {
```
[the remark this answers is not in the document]
nimakarimipour: b6e4a1192cb8a7b4f17f0de56c8dfa3c7878e4fa and 795c25d87e2c48e06028bdb1d822d14873217727.

Yes that is exactly what happens. Regarding `MethodDeclarationTree` we had a PR to rename that #54 but I agree that still it is not the best name. One thing I have in my mind to do in future is to combine both `fieldDeclarationStore` and `MethodDeclarationStore` to a single class containing all knowledge regarding declaration of elements in source code.

nimakarimipour:
@msridhar @lazaroclapp Thank you for the review, this is ready for another round. Thank you.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -116,21 +118,22 @@ private void preprocess() {
 
   /** Performs iterations of inference/injection until no unseen fix is suggested. */
   private void annotate() {
-    // globalAnalyzer analyzes effects of all public APIs on downstream dependencies.
+    // Downstream impact cache stores impact of making all public APIs @Nullable on downstream
```
msridhar, line 121:
```suggestion
    // The downstream impact cache stores the impact of making all public APIs @Nullable on downstream
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/Impact.java
```diff
@@ -0,0 +1,97 @@
...
+
+/** Stores the set of errors which will be triggered if the containing fix is applied. */
+public class Impact {
+
+  /** Target fix. */
+  public final Fix fix;
+  /** Set of triggered errors, if this fix is applied to source code. */
+  protected ImmutableSet<Error> triggeredErrors;
+  /** Set of triggered fixes on downstream dependencies if containing fix is applied. */
+  protected ImmutableSet<Fix> triggeredFixesOnDownstream;
```
msridhar, line 43: Why do we say triggered fixes _on_ downstream dependencies?  I thought we cannot apply fixes to downstream dependencies, only the target?

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/Impact.java
```diff
@@ -0,0 +1,97 @@
...
+   *
+   * @param fixes Set of applied fixes to source code permanently.
+   */
+  public void updateStatusAfterInjection(Collection<Fix> fixes) {
+    triggeredErrors =
+        ImmutableSet.copyOf(
+            triggeredErrors.stream()
+                .filter(error -> !error.isResolvableWith(fixes))
+                .collect(Collectors.toSet()));
+  }
```
msridhar, line 69: Instead of `copyOf()` you can use the `Collector` returned by [`toImmutableSet()`](https://guava.dev/releases/31.0-jre/api/docs/com/google/common/collect/ImmutableSet.html#toImmutableSet())

msridhar (commented):
Not at all a complete review, just a question and couple of quick comments

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/Impact.java
msridhar: The comment in `Report.java` is clearer:

https://github.com/ucr-riple/NullAwayAnnotator/blob/0f1ca3271126ace7500e83cdeb6c9ef4d28515da/annotator-core/src/main/java/edu/ucr/cs/riple/core/Report.java#L55-L59

Can we rename these fields to `triggeredFixesFromDownstreamErrors` for clarity?

nimakarimipour: No, we are not applying any fix to downstream dependencies, renamed for clarity, thank you for the suggestion. c4100c0fe640fe656945968b720720f98b799b99

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/Impact.java
nimakarimipour: 79fb387a8d2c472e11af301e242d5e29b854f31f

nimakarimipour:
@msridhar @lazaroclapp This is ready for another review. Thank you very much.

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpact.java
```diff
@@ -24,59 +24,51 @@
...
 import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-/** Container class for storing overall effect of each method. */
-public class DownstreamImpact {
-  /** Node in {@link MethodDeclarationTree} corresponding to a public method. */
-  final MethodNode node;
+/**
+ * Container class for storing overall effect of fix in downstream dependencies. At this moment,
```
lazaroclapp, line 44: Question: is this really about the impact of a fix **in** downstream dependencies (i.e. is the fix applied to the code of downstream dependencies) or is this the impact of a fix in the target **on** the downstream dependencies (errors that show up when building downstream dependencies due to a change in the code of the target)

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpact.java
```diff
@@ -24,59 +24,51 @@
...
    * dependencies.
    */
   private int effect;
 
-  public DownstreamImpact(MethodNode node) {
-    this.node = node;
+  public DownstreamImpact(Fix fix) {
+    super(fix);
+    // Only store impacts of fixes targeting methods.
+    Preconditions.checkArgument(fix.isOnMethod());
```
lazaroclapp, line 63:
```suggestion
    Preconditions.checkArgument(fix.isOnMethod(), "Unexpected Fix instance. Only impacts of fixes on methods should be tracked for downstream dependencies");
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpactCacheImpl.java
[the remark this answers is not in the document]
lazaroclapp: Those two commits are not in this PR...

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/suppliers/DownstreamDependencySupplier.java
```diff
@@ -74,8 +75,14 @@ public DownstreamImpactCache getDownstreamImpactCache() {
...
-    return config.optimized
+    return config.useParallelGraphProcessor
         ? new ParallelConflictGraphProcessor(config, runner, this, tracker)
         : new SequentialConflictGraphProcessor(config, runner, this);
   }
+
+  @Override
+  public TargetModuleCache getTargetModuleCache() {
+    throw new RuntimeException(
+        "Caches are used to retrieve impacts for depths more than 1. Downstream dependency analysis happens only at depth 1.");
```
lazaroclapp, line 86: I don't understand this error message. Is this really related to analysis depth? I'd think the issue here is that we are asking for the `TargetModuleCache` when running in a mode that is only configured to use the cache for downstream dependencies. So this is an `UnsupportedOperationException` and the message needs to be changed. Or am I getting something wrong?

lazaroclapp (commented):
Did a pass from scratch to the best of my ability (since by now I have lost context from our previous rounds of reviews). I think the main question is around whether the cache should be turned by default and whether or not we have tests for it.

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/CoreTestHelper.java
```diff
@@ -280,11 +280,12 @@ public void makeAnnotatorConfigFile(Path configPath) {
     builder.redirectBuildOutputToStdErr = true;
     builder.chain = true;
     builder.outerLoopActivation = requestCompleteLoop;
-    builder.optimized = true;
+    builder.useParallelProcessor = true;
     builder.downStreamDependenciesAnalysisActivated = downstreamDependencyAnalysisActivated;
     builder.mode = mode;
     builder.inferenceActivated = !deactivateInference;
     builder.forceResolveActivation = forceResolveActivated;
+    builder.useCacheImpact = false;
```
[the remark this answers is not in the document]
nimakarimipour: I changed the default configuration to disable caching. 099b6d4e246afd6157817c69e1b5092eb55c434d

In #127 we update CI to run all unit tests in all four combinations. This also revealed a problem and got fixed. I think we can enable caching for this PR by default for unit tests for now to make sure all tests pass and then land #127 which runs all tests in all configurations. Please let me know if you agree this. Thank you.

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -436,4 +437,62 @@ public void staticBlockLocalVariableInitializationTest() {
...
+        .toDepth(5)
+        .disableBailOut()
+        .addExpectedReports(
+            new TReport(
+                new OnMethod("A.java", "test.A", "foo()"),
+                -2,
+                Set.of(new OnField("A.java", "test.A", Collections.singleton("f"))),
+                Collections.emptySet()),
+            new TReport(new OnField("A.java", "test.A", Collections.singleton("f")), -1))
+        .setPredicate((expected, found) -> expected.testEquals(coreTestHelper.getConfig(), found))
```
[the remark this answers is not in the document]
nimakarimipour: @lazaroclapp Yes excatly that is correct. I will update the javadoc in a followup PR to make this clearer.

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/Impact.java
nimakarimipour: Sorry, happened in another branch ce46c8ffb53747f41330ad098b550f37fa21f90b

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/BaseCache.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Override
+  public boolean isUnknown(Fix fix) {
+    return !this.store.containsKey(fix.toLocation());
+  }
+
+  @Nullable
+  @Override
+  public T fetchImpact(Fix fix) {
+    return store.getOrDefault(fix.toLocation(), null);
```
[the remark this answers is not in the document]
nimakarimipour: Thought it might throw exception as key not found (confused it with python dicts) but reading the javadoc it does not. 9ec4550fb3ee0962c445efaa904c72e49f8a5db0

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpact.java
nimakarimipour: Yes this was not clear. It is the second case, the impact of a fix in the target on the downstream dependencies. Updated jvadoc clarity. thank you for noticing this. 21e7cefe122a7799ed6fa1b9c0d948f72ab7e215

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/ErrorStore.java
```diff
@@ -104,4 +105,11 @@ public Result compare() {
   public Set<Region> getRegionsForElements(Predicate<Error> predicate) {
     return root.getRegionsOfMatchingItems(predicate);
   }
+
+  public int getNumberOfResolvedFixesWithCollection(Collection<Fix> fixes) {
```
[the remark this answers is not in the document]
nimakarimipour: fixed in 73009539bfc6e61b4fc95c90af1e3035dea16974

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/ErrorStore.java
```diff
@@ -104,4 +105,11 @@ public Result compare() {
...
+    return root.values().stream()
+        .filter(error -> error.isResolvableWith(fixes))
+        .collect(Collectors.toSet())
+        .size();
```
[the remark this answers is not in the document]
nimakarimipour: fixed in 73009539bfc6e61b4fc95c90af1e3035dea16974

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -436,4 +437,62 @@ public void staticBlockLocalVariableInitializationTest() {
         .enableForceResolve()
         .start();
   }
+
+  @Test
```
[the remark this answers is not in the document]
nimakarimipour: Yes, I agree that these two unit tests are not directly related, but they helped me discover a bug in caching. I extracted into a separate PR that only adds these two tests. #138

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/FieldRegionTracker.java
```diff
@@ -71,6 +71,12 @@ public Optional<Set<Region>> getRegions(Location location) {
             .map(trackerNode -> trackerNode.region)
             .collect(Collectors.toSet());
     ans.addAll(config.getAdapter().getFieldRegionScope(field));
+    if (!config.useImpactCache) {
+      // Errors of type `METHOD_NO_INIT` and `FIELD_NO_INIT` are reported in null region that might
+      // have been resolved. Caching approach does not rely on build results to compute the set of
+      // resolved errors.
```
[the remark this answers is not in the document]
nimakarimipour: This was necessary for NullAway `0.10.6`. Since in this PR NullAway 0.10.8 is used, we no longer require that and is removed now. This was regarding the problem with serialization for initialization errors in 0.10.6.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -235,11 +237,18 @@ public Config(String[] args) {
...
+        new Option(
+            "dpp", "disable-parallel-processing", false, "Disables parallel processing of fixes");
+    disableParallelProcessingOption.setRequired(false);
+    options.addOption(disableParallelProcessingOption);
+
+    // Fix impact cache
+    Option disableFixCacheOption =
+        new Option("dfc", "disable-fix-cache", false, "Disables fix impact caching");
+    disableFixCacheOption.setRequired(false);
+    options.addOption(disableFixCacheOption);
```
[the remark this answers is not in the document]
nimakarimipour: Yes I agree that The option should be and is turned off by default now.

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/suppliers/DownstreamDependencySupplier.java
nimakarimipour: Impacts of fixes in target module on downstream dependencies are computed via `DownstreamImpactEvaluator`. `DownstreamImpactEvaluator` is a specialized instance of `BasicEvaluator` that from point of view of `DownstreamImpactEvaluator`, downstream dependencies of target module is the target module. Caching can have use case only if the depth of search is greater than one. Therefore, if the target module cache is acquired by the Downstream dependency evaluator, the search on downstream dependencies is happening for depth greater than 1 which as of this moment we are not supporting (it can be done with current implementation but just prohibited). Please let me know if I should change the exception message. Thank you.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/ParameterRegionTracker.java
```diff
@@ -62,7 +62,8 @@ public Optional<Set<Region>> getRegions(Location location) {
             .collect(Collectors.toSet());
     // Add the method the fix is targeting.
     regions.add(new Region(parameter.clazz, parameter.method));
-    // Add all call sites.
+    // Add all call sites. It will also reserve call sites to prevent callees have their parameters
+    // annotated simultaneously. See test: CoreTest#nestedParameters.
```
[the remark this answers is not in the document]
nimakarimipour: I extracted the test case and the comment to #138. The main issue is that, I wanted to emphasize for parameter, the call sites of the method, can be a potentially impacted region.

Just to avoid scenario below:

```java
void foo(Object p){
       bar(p);
}

void bar(Object p)
```

assume that we have two fixes on foo#bar and `bar#p` and `foo#p`, while we are analyzing impact of making `foo#p` nullable, `bar#p` should not be annotated as nullable, otherwise we will miss the triggered error that p is nullable while invoking `bar`. This addition in effective region will make a region conflict for `foo#p` and `bar#p` and they will not be processed simultaneously.

annotator-core/src/main/java/edu/ucr/cs/riple/core/evaluators/suppliers/DownstreamDependencySupplier.java
lazaroclapp: I am still not understanding the explanation here. (What I am sort of getting is that "depth" here is used in a different meaning from the already existing concept of "depth" in the auto-annotator, no? i.e. this isn't depth of the chain of fixes, this is depth of the build graph, which only adds to the confusion. Or am I get that part wrong too?)

lazaroclapp (approved):
One small nit and I am still a bit confused by the wording on that one exception message. That said, at this point I am pretty convinced of the code of this PR and discussion is growing unwieldy. Happy to see this landed with the javadoc nit below (please double-check formatting), and have the exception message discussion moved to a follow up issue/PR!

PR #112, opened 2022-12-14
Enhance error data structure.

nimakarimipour:
This PR is build upon #103.

In NullAway, if the error is resolvable and involved a pseudo-assignment of a `@Nullable` expression into a `@NonNull` target, the location of that target is serialized. This information is enough to generate a fix resolving the reported error. Currently we only store the location of the `@Nonnull` element in `Error` data structure. This PR updates `Error` data structure to store the actual fix which can resolve the error.

Changes in this PR
- Update Error data structure to store the resolving fixes.
- Creates fixes according to the information serialized for the reporting error.

PR #113, opened 2022-12-14
Remove triggered fixes from iterations.

nimakarimipour:
#### ~~This PR is built upon #112.~~ #112 is landed.

Following #112 errors contain the set of fixes which can resolve the reported errors. Currently, after each fix examination, the set of triggered fixes and errors are computed per region. Since #112 updates error instances to store the resolving fixes, the set of triggered fixes (which resolves the triggered errors) is redundant. This PR removes this structure.

Changes in this PR:

- Removes set of triggered fixes from `Nodes` in `ConflictGraph`.
- Constructs fix trees based on the containing resolving fixes of triggered errors.
- Adds set of triggered fixes from downstream to `nodes` for clear separation between fixes from downstream and fixes from target.


annotator-core/build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
 }
 
 // Should be the latest supporting version of NullAway.
-def NULLAWAY_TEST = "0.10.5"
+def NULLAWAY_TEST = "0.10.6-SNAPSHOT"
```
msridhar, line 53: I guess we need to wait on a new NullAway release before landing this one?

annotator-core/src/main/java/edu/ucr/cs/riple/core/Report.java
```diff
@@ -46,17 +46,18 @@ public class Report {
...
   public ImmutableSet<Error> triggeredErrors;
-  /** If true, all leaves of fix tree are not resolvable by any {@code @Nullable} annotation. */
-  public boolean finished;
+  /**
+   * Set of triggered fixes on target module that will be triggered if fix tree is applied due to
+   * errors in downstream dependencies.
+   */
+  public ImmutableSet<Fix> triggeredFixesOnDownstream;
+  /** If true, this report has passed one iteration */
+  public boolean opened;
```
msridhar, line 60: Why call this `opened`?

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/TError.java
```diff
@@ -0,0 +1,36 @@
...
+ * THE SOFTWARE.
+ */
+
+package edu.ucr.cs.riple.core.tools;
+
+import edu.ucr.cs.riple.core.metadata.index.Error;
+import edu.ucr.cs.riple.core.metadata.trackers.Region;
+import edu.ucr.cs.riple.injector.location.Location;
+
+public class TError extends Error {
```
msridhar, line 31: What is the purpose of this class?  Maybe give a clearer name?

msridhar (commented):
Not an expert on this code so the review was rather superficial.

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/TError.java
lazaroclapp: Worth documenting. However, if we want to rename this to something like `ErrorMatcher` or `ExpectedError`, we probably want to do so in a different diff, as this `T[X]` pattern is common for these classes to match a "expected X" in tests.

annotator-core/src/test/java/edu/ucr/cs/riple/core/DeepTest.java
```diff
@@ -87,6 +87,39 @@ public void field_assign_nullable_bailout_disabled() {
         .start();
   }
 
+  @Test
+  public void testTwoReports() {
```
lazaroclapp, line 91: What are we testing here that wasn't the case before this PR? I am not sure I follow.

annotator-core/build.gradle
nimakarimipour: Yes, we do. Please let me know if you would like me to update CI or gradle scripts to use locally snapshots for CI to pass for now. Thank you.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Report.java
nimakarimipour: I wanted to distinguish between a report object that has just been created and one that has been processed for at least one iteration. Renamed it to `hasBeenProcessedOnce`.  2bdd664ec26ba0d5536ac345547550be26698254.

I didn't want to create a separate enum for this as it only has two cases and can be implemented with a single boolean. But if you think enum is a better design here, please let me know and I will update it. Thank you.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Report.java
```diff
@@ -244,7 +253,19 @@ public int getUpperBoundEffectOnDownstreamDependencies() {
...
    */
   public boolean isInProgress(Config config) {
-    return (!finished && (!config.bailout || localEffect > 0))
-        || triggeredFixes.stream().anyMatch(input -> !input.fixSourceIsInTarget);
+    if (!opened) {
+      // report has not been processed.
+      return true;
+    }
+    if (triggeredFixesOnDownstream.size() != 0 && !tree.containsAll(triggeredFixesOnDownstream)) {
+      // force to processes move forward with the triggered fix in downstream dependencies.
```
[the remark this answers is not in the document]
nimakarimipour: 3623359a9a7811a2e7348f683ae378c3c154bc97

Sorry it wasn't clear. Updated comment.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -78,6 +86,19 @@ public static Factory<Error> factory(Config config, FieldDeclarationStore store)
...
+   * Checks if error is resolvable.
+   *
+   * @return true if error is resolvable and false otherwise.
+   */
+  public boolean hasFix() {
+    return this.resolvingFixes.size() > 0;
+  }
+
+  public ImmutableSet<Fix> getResolvingFixes() {
+    return ImmutableSet.copyOf(this.resolvingFixes);
```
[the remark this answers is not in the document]
nimakarimipour: 391c08615c29d178443ec1bec4ce70a0775faea3

annotator-core/src/test/java/edu/ucr/cs/riple/core/DeepTest.java
nimakarimipour: This is specifically designed to make sure we do not count an error for a constructor failing to initialize `@Nonnull` fields multiple times.

In the example below and in the unit test:
```java
class Main{
      Object field1;
      Object field2;
      Main() {
          field1 = null;
          field2 = null;
      }
}
```
Here we have an error from the constructor that is failing to initialize `field1` and `field2`. If we annotate `field1` as `@Nullable` we will still get the error that constructor is failing to initialize `field2`. We shouldnt double count this as a new error and include its resolving fixes in the tree.


Renamed unit test f8946f4abb76f65e087198648fd11e8d16dcbb84

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/TError.java
nimakarimipour: @msridhar @lazaroclapp I am assuming that the decision here it to leave this as is with added comment 4608ac336c4450cb2243eed068a5ef328c53fa28.

nimakarimipour:
@msridhar @lazaroclapp Thank you very much for your comments. This PR is ready for another round of review.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -87,7 +92,7 @@ public static Factory<Error> factory(Config config, FieldDeclarationStore store)
   }
 
   /**
-   * Checks if error is resolvable.
+   * <<<<<<< HEAD Checks if error is resolvable.
```
lazaroclapp, line 95: Again, rebase issue here!

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -96,11 +101,11 @@ public boolean hasFix() {
   }
 
   public ImmutableSet<Fix> getResolvingFixes() {
-    return ImmutableSet.copyOf(this.resolvingFixes);
+    return this.resolvingFixes;
   }
 
   /**
-   * Checks if error is resolvable with only one fix.
+   * ======= >>>>>>> master Checks if error is resolvable with only one fix.
```
lazaroclapp, line 108: And here!

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/TError.java
```diff
@@ -28,6 +28,7 @@
 import edu.ucr.cs.riple.core.metadata.trackers.Region;
 import edu.ucr.cs.riple.injector.location.Location;
 
+/** Wrapper class for {@link Error} with default values, used to create tests. */
```
lazaroclapp, line 31:
```suggestion
/** Wrapper class for {@link Error} used to create dummy errors (with default values) as part of {@link TReport} matchers, which in turn are used by tests to define expected reports to be matched by the actual reports produced by the test case. */
```

Something like that? (Assuming GJF takes care of making it multi-line). 

Probably worth updating the javadocs for `TFix` and `TReport` as well. 

In general, when writing a javadoc, don't just write something that satisfies the requirement of "there is now a javadoc comment here". Instead, it's a good idea to try to make them informative for a reader that doesn't have the full tool's architecture in mind at the time. Not always possible, but a good question to ask is: if someone landed on the repo by way of this class, how useful is this javadoc in explaining the role the class/method fulfills in general or, failing that, pointing them to a place where they can get a better understanding of how it's used? The original "to create tests" phrasing tells someone this is used for testing, but... how? As a mock object? As something reported from the test? Something else? I _think_ the answer is what I am saying above "to construct a matcher", but that isn't obvious without reading the code for `TReport` and then either looking at the test cases or maybe at `CoreTestHelper`.

lazaroclapp (commented):
I have probably lost enough context on this PR that I might need to do another pass from the beginning, but for now, just a couple quick comments on the existing relevant discussion threads and changes since last review :) 

Also, tests are failing... incomplete rebase / rebase issues? Or is it just that this one is blocked on a new NullAway version either way?

annotator-core/build.gradle
msridhar: I don't think we should use snapshots in CI.  Is the idea that we will only land this change after cutting a new NullAway release and updating?

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -166,4 +187,19 @@ public int hashCode() {
...
+   * @param errors Collection of errors.
+   * @return Immutable set of fixes which can resolve all given errors.
+   */
+  public static ImmutableSet<Fix> getResolvingFixesOfErrors(Collection<Error> errors) {
+    Map<Fix, Set<Set<String>>> m =
+        errors.stream()
+            .flatMap(error -> error.resolvingFixes.stream())
+            .collect(groupingBy(identity(), mapping(fix -> fix.reasons, Collectors.toSet())));
+    m.forEach((fix, sets) -> sets.forEach(fix.reasons::addAll));
+    return ImmutableSet.copyOf(m.keySet());
```
msridhar, line 203: FWIW, I find this stream-based code to be somewhat unreadable.  I guess line 202 is doing some kind of mutation on the fixes?  Probably this one would be clearer with loops and iterators.

msridhar (commented):
I think I followed the high-level idea here, though not all the details

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
nimakarimipour: ba333098e865557891a1fbf40e49e18f24ae4e75

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/TError.java
nimakarimipour: Thank you for the notes. I did not have in my mind these different use cases of helpers in tests you mentioned, and the original javadoc was definitely incomplete and vague. Updated the rest as well fcdf38eac7b229d4f3029449d66847c174b4ff42, will have this in my mind for future PRs.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
nimakarimipour: cae2f74cea9219a4916a54f8fe2c094a4c619fd7

Thank you for the suggestion. It was mutating but wasn't making any problems, however, I rewrote to be more cautious.

nimakarimipour:
> I have probably lost enough context on this PR that I might need to do another pass from the beginning, but for now, just a couple quick comments on the existing relevant discussion threads and changes since last review :)
> 
> Also, tests are failing... incomplete rebase / rebase issues? Or is it just that this one is blocked on a new NullAway version either way?

@lazaroclapp CI is blocked by the new NullAway release which includes [#704](https://github.com/uber/NullAway/pull/704)

annotator-core/build.gradle
nimakarimipour: @msridhar Yes, we can land it once the new NullAway releases. I will keep checking status of changes locally.

nimakarimipour:
@msridhar @lazaroclapp Thank you for the review, this is ready for another round.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Fix.java
```diff
@@ -24,72 +24,42 @@
...
  * are coming form NullAway.
  */
-public class Fix extends Enclosed {
+public class Fix {
 
   /** Suggested change. */
   public final AddAnnotation change;
   /** Reasons this fix is suggested by NullAway in string. */
-  public final Set<String> reasons;
+  public final ImmutableSet<String> reasons;
```
msridhar, line 47: Eventually (not in this PR), maybe we could add an enum `FixReason` and then make this an `ImmutableSet<FixReason>`.  My guess is there is only a small set of possible reasons

msridhar (approved):
From my end this looks good

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Fix.java
nimakarimipour: @msridhar Sure, that sounds good. Long time ago I had a dependency to NullAway and used the `ErrorMessage.MessageTypes` enum. But I removed it since we are adding a dependency for just an enum. Would you please tell me if I should still avoid adding a dependency and create a our own `FixReason` considering that they are identical ?

annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/TReport.java
```diff
@@ -34,7 +34,11 @@
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
-/** Wrapper class for {@link Report} with default values, used to create tests. */
+/**
+ * Wrapper class for {@link Report} used to create dummy reports (with default values) which in turn
```
lazaroclapp, line 38:
```suggestion
 * Wrapper class for {@link Report} used to create dummy reports (with default values) which
```

No need for this "in turn" here, since the reports are just directly used by test to [...] :)

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Fix.java
lazaroclapp: I think it is indeed worth it to have our own copy of `MessageTypes` or `FixReason` and keep it updated. I suppose the ideal answer here is for there to be a `com.uber.nullaway:tooling-lib:[...]` or something that could be depended on instead of the main annotation processor, but that's not likely to happen immediately (we still don't ship a user library with e.g. `@Initializer` so...).

That said, definitely a thought for a different - very low priority - PR. Might not even be worth it before the current release for the auto-annotator. @msridhar ?

lazaroclapp (commented):
Recent changes all look good to me, with a minor nit on a comment! 

Might want to do a super quick pass over the full logic once we have the new NullAway release, since we need that to run CI/land this, right?

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Fix.java
msridhar: Agree this is low priority and we can just track with an issue. I also think it's ok to just copy the enum and manually sync for now.  But it will get updated as and when we add more JSpecify generics support

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -525,21 +523,19 @@ public void initializeAdapter(FieldDeclarationStore fieldDeclarationStore) {
     Path serializationVersionPath = target.dir.resolve("serialization_version.txt");
     if (!serializationVersionPath.toFile().exists()) {
       // Older versions of NullAway.
-      this.adapter = new NullAwayV0Adapter(this, fieldDeclarationStore);
-      return;
+      throw new RuntimeException(
+          "This annotator version does not support using version of NullAway, please upgrade NullAway or use version 1.3.4 of Annotator.");
```
lazaroclapp, line 527: Can we specify _which_ versions of NullAway are (un)supported?
```suggestion
          "This annotator version requires NullAway 0.10.6 or higher (using SerializeFixMetadataVersion=2), please upgrade NullAway or use version 1.3.4 of Annotator.");
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -525,21 +523,19 @@ public void initializeAdapter(FieldDeclarationStore fieldDeclarationStore) {
...
     try {
       List<String> lines = Files.readAllLines(serializationVersionPath);
       int version = Integer.parseInt(lines.get(0));
       switch (version) {
         case 0:
-          this.adapter = new NullAwayV0Adapter(this, fieldDeclarationStore);
-          this.offsetHandlingIsActivated = false;
-          break;
+          throw new RuntimeException(
+              "This annotator version does not support serialization version 0, please upgrade NullAway or use version 1.3.5 of Annotator.");
```
lazaroclapp, line 535:
```suggestion
              "This annotator version does not support serialization version 0, please upgrade NullAway to 0.10.6+ (with SerializeFixMetadataVersion=2) or use version 1.3.5 of Annotator.");
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -525,21 +523,19 @@ public void initializeAdapter(FieldDeclarationStore fieldDeclarationStore) {
...
         case 1:
-          this.adapter = new NullAwayV1Adapter(this, fieldDeclarationStore);
-          this.offsetHandlingIsActivated = false;
-          break;
+          throw new RuntimeException(
+              "This annotator version does not support serialization version 1, please upgrade NullAway or use version 1.3.5 of Annotator.");
```
lazaroclapp, line 538:
```suggestion
              "This annotator version does not support serialization version 1, please upgrade NullAway to 0.10.6+ (with SerializeFixMetadataVersion=2) or use version 1.3.5 of Annotator.");
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -166,4 +184,39 @@ public int hashCode() {
...
+              if (fixReasonsMap.containsKey(fix)) {
+                fixReasonsMap.get(fix).addAll(fix.reasons);
+              } else {
+                fixReasonsMap.put(fix, new HashSet<>(fix.reasons));
+              }
+            });
+
+    ImmutableSet.Builder<Fix> builder = ImmutableSet.builder();
+    for (Fix key : fixReasonsMap.keySet()) {
+      // To avoid mutating fixes stored in the given collection, we create new instances.
```
lazaroclapp, line 215:
```suggestion
      // To avoid mutating fixes stored in the given collection, we create new instances.
      // which contain the full set of reasons.
```

annotator-core/src/test/java/edu/ucr/cs/riple/core/DeepTest.java
```diff
@@ -87,6 +87,39 @@ public void field_assign_nullable_bailout_disabled() {
...
+            "Main.java",
+            "package test;",
+            "public class Main {",
+            "   Object field1;",
+            "   Object field2;",
+            "   Main() {",
+            "     field1 = null;",
+            "     field2 = null;",
+            "   }",
+            "   Object getF1(){ return field1;}",
```
lazaroclapp, line 103:
```suggestion
            "   Object getF1(){ return field1; }",
```

annotator-core/src/test/java/edu/ucr/cs/riple/core/DeepTest.java
```diff
@@ -87,6 +87,39 @@ public void field_assign_nullable_bailout_disabled() {
...
+            "   Object getF2(){ return field2;}",
```
lazaroclapp, line 104:
```suggestion
            "   Object getF2(){ return field2; }",
```

lazaroclapp (approved):
So, it's a large PR and I could be missing something, but from a quick re-review, it LGTM!

There are three suggestions below, but they are trivial (and optional) nits! Happy New Year, Nima!

msridhar (approved):

PR #116, opened 2022-12-20
Extract AnnotatorScanner config writer into a separate class.

nimakarimipour:
This PR simply extracts the config writer inside the `annotator-scanner` module into a separate standalone class, and reuses it in `annotator-core` module.

Please note that this is a preparation for the upcoming PR handling region selection for generated code.


msridhar (approved):

nimakarimipour:
@msridhar Thank you for the review, I am going to land this now.

PR #118, opened 2022-12-20
Include generated code in determining effective regions.

nimakarimipour:
This PR is built on #116

This PR enhances logic for selecting the potentially impacted regions to include generated code by processors (e.g. [Lombok](https://projectlombok.org)).

`Trackers` are used to detect potentially impacted regions for a fix. These trackers only include the regions where the element was used directly. But since some processors, will copy the annotations on the element over some generated methods, these trackers will miss the real impacted regions because of the copied annotation. This PR tries to address this issue.

Changes in this PR:
- Adds `GeneratedCodeDetector` interface in `annotator-scanner` module which is responsible for determining if a `TreePath` is pointing to a generated code.
- Adds `LombokGeneratedCodeDetector`  which implements `GeneratedCodeDetector` and can detect if a path is pointing to a `Lombok` generated code. For this feature to work properly (`lombok.addLombokGeneratedAnnotation = true` must be set)
- Adds a fifth column in the element use serialization in regions which denotes if the serialized region exists in source code or is generated by a processor. If the region exists in the source code it will have the value `"SOURCE"`, otherwise, it will have the name of the processor that created it. 
- Adds `GeneratedRegionTracker` in `annotator-core` module which is responsible for extending the list of potentially impacted regions to include users of generated code that uses the target element.
- Adds `LombokTracker` which implements `GeneratedRegionTracker` and can extend regions with other regions that Lombok has created.

To enable handling regions created by Lombok, the flag below must be passed:
```
-ardl, --activate-region-detection-lombok
```


annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
```diff
@@ -72,4 +73,14 @@ public interface Config {
...
+
+  /**
+   * Returns source for the element at the given path. If the element exists in the source code
+   * {@code "SOURCE"} will be returned and otherwise the name of the code generator will be
+   * returned.
+   *
+   * @return Name of the generator that produced the code and {@code "SOURCE"} if element exists in
+   *     source code.
+   */
+  String getSourceForSymbolAtPath(TreePath path);
```
msridhar, line 85: Again use enum for return

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/LombokGeneratedCodeDetector.java
```diff
@@ -0,0 +1,67 @@
...
+ * Detector for generated code by <a href="https://projectlombok.org">Lombok</a>. The basic
+ * assumption used here is that all generated code by Lombok is inside a method annotated with <a
+ * href="https://projectlombok.org/api/lombok/Generated">lombok.Generated</a> annotation. Please
+ * note that {@code "lombok.addLombokGeneratedAnnotation = true"} should be set in lombok
+ * configurations while building target module to add the required annotations.
+ */
+public class LombokGeneratedCodeDetector implements GeneratedCodeDetector {
+
+  /** Associated name for Lombok in region serializations in <i>SOURCE_TYPE</i> column. */
+  private static final String NAME = "LOMBOK";
```
msridhar, line 43: use enum

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
```diff
@@ -0,0 +1,41 @@
...
+  public LombokTracker(
+      MethodDeclarationTree methodDeclarationTree, MethodRegionTracker methodRegionTracker) {
+    this.methodDeclarationTree = methodDeclarationTree;
+    this.tracker = methodRegionTracker;
+  }
+
+  @Override
+  public Set<Region> extendWithGeneratedRegions(Set<Region> regions) {
+    return regions.stream()
+        // filter regions where are created by lombok
```
msridhar, line 26: I think the idea here is that for any Lombok-generated method in `regions`, we also add all _uses_ of that method to the impacted regions.  Is this the right idea?  If so, let's add some comments / Javadoc documenting the logic here.  Also we should state briefly why we think this is sufficient to get all the relevant regions, given the documented assumptions on how Lombok propagates annotations (see previous comment)

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
```diff
@@ -0,0 +1,41 @@
...
+import edu.ucr.cs.riple.core.metadata.method.MethodDeclarationTree;
+import edu.ucr.cs.riple.core.metadata.trackers.MethodRegionTracker;
+import edu.ucr.cs.riple.core.metadata.trackers.Region;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class LombokTracker implements GeneratedRegionTracker {
```
msridhar, line 12: Let's add some Javadoc on this class that documents what our assumptions are in terms of how exactly Lombok propagates annotations, so we know what we are trying to handle.

msridhar (commented):
I think I get the high-level idea.  I have a few comments, mostly around use of `enum`s instead of `String`s

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -141,6 +141,8 @@ public class Config {
    */
   private NullAwayVersionAdapter adapter;
 
+  public final ImmutableSet<String> generatedCodeDetectors;
```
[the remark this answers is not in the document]
nimakarimipour: Thank you for the suggestion, I used enum 4bbd2e4c7539c60756522d7238db930d34bcd1d2

annotator-core/src/main/java/edu/ucr/cs/riple/core/adapters/NullAwayV0Adapter.java
```diff
@@ -87,12 +87,12 @@ public Error deserializeError(String[] values) {
...
     Preconditions.checkArgument(
-        values.length == 4,
-        "Expected 4 values to create TrackerNode instance in NullAway serialization version 0 but found: "
+        values.length == 5,
+        "Expected 5 values to create TrackerNode instance in NullAway serialization version 0 but found: "
             + values.length);
     String regionMember =
         !Region.getType(values[1]).equals(Region.Type.METHOD) ? "null" : values[1];
-    return new TrackerNode(new Region(values[0], regionMember), values[2], values[3]);
+    return new TrackerNode(new Region(values[0], regionMember, values[4]), values[2], values[3]);
```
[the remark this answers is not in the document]
nimakarimipour: Yes exactly.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/Region.java
```diff
@@ -38,6 +38,11 @@ public class Region {
   public final String member;
   /** Fully qualified name of the enclosing class of the region. */
   public final String clazz;
+  /**
+   * Type of region. If region exists in source code, this value is {@code "SOURCE"}, otherwise it
+   * will be the name of the processor created this region. (e.g. {"LOMBOK"}).
+   */
+  public final String sourceType;
```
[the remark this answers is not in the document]
nimakarimipour: 4bbd2e4c7539c60756522d7238db930d34bcd1d2

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
nimakarimipour: 4bbd2e4c7539c60756522d7238db930d34bcd1d2

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/LombokGeneratedCodeDetector.java
nimakarimipour: 4bbd2e4c7539c60756522d7238db930d34bcd1d2

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
nimakarimipour: f0880e2067988c53d64832202e269c34e7366588

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
nimakarimipour: f0880e2067988c53d64832202e269c34e7366588

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
```diff
@@ -0,0 +1,77 @@
...
+        .filter(Objects::nonNull)
+        // get method location.
+        .map(methodNode -> methodNode.location)
+        // add potentially impacted regions for the collected methods.
+        .flatMap(
+            onMethod -> {
+              Optional<Set<Region>> ans = tracker.getRegions(onMethod);
+              return ans.isPresent() ? ans.get().stream() : Stream.of();
+            })
+        .collect(Collectors.toSet());
```
lazaroclapp, line 75: This works on methods only and one hop, right? I.e. if our normal exploration adds `@Nullable` to field `Foo.bar` and `Foo` is `@Data`, our initial affected regions (_before_ this fix) will include `getBar()`. Then this chance makes sure we add all methods/fields/etc that would be affected by `getBar()` itself automatically becoming `@Nullable` due to Lombok. Correct?

This might not be enough to track all the changes to fields and methods of the inner class `Foo.Builder` or their impacted regions if the class is also annotated `@Builder`. Right?

Happy if the answer is no and we can do that other case as a _follow up_ PR, btw 🙂

lazaroclapp (commented):
Did a pass. Also, again, I think the description could use a bit of an edit/proofread. It's useful, but there are a few sentences that are harder to parse than they need to be.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/Config.java
```diff
@@ -72,4 +74,14 @@ public interface Config {
...
+
+  /**
+   * Returns source for the element at the given path. If the element exists in the source code
+   * {@code "SOURCE"} will be returned and otherwise the name of the code generator will be
+   * returned.
+   *
+   * @return Associated Source type of the generator that produced the code. If element exists in
+   *     source code and not produced by any processor, it will return {@link SourceType#SOURCE}
+   */
+  SourceType getSourceForSymbolAtPath(TreePath path);
```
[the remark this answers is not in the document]
nimakarimipour: 10559d1ee17a69f81b170f6f3179084ea9d9fd65

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/LombokGeneratedCodeDetector.java
```diff
@@ -0,0 +1,67 @@
...
+  /** Associated source type for Lombok in region serializations in <i>SOURCE_TYPE</i> column. */
+  private static final SourceType sourceType = SourceType.LOMBOK;
+
+  @Override
+  public boolean isGeneratedCode(TreePath path) {
+    MethodTree enclosingMethod =
+        path.getLeaf() instanceof MethodTree
+            ? (MethodTree) path.getLeaf()
+            : ASTHelpers.findEnclosingNode(path, MethodTree.class);
+    if (enclosingMethod == null && path.getLeaf() instanceof MethodTree) {
```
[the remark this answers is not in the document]
nimakarimipour: 5a0eee109f07d37d5c2b993cb84f47a5ebad38c6

Yes, that it true, thank you for noticing this.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/LombokGeneratedCodeDetector.java
```diff
@@ -0,0 +1,67 @@
...
+      enclosingMethod = (MethodTree) path.getLeaf();
+    }
+    if (enclosingMethod == null) {
+      return false;
```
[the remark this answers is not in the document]
nimakarimipour: I was only focusing on generated methods which can be called outside of the class and extend the potentially impacted regions. But, you are right, I don't see any harm in having a comprehensive check on the generated code of any type. Please see 3621febe383f74896a79cb75ad777a417bd25504. Used the same logic for detecting region member which already covers all cases.

annotator-core/src/test/resources/templates/lombok/build.gradle
```diff
@@ -0,0 +1,76 @@
...
+tasks.withType(JavaCompile) {
+    // remove the if condition if you want to run NullAway on test code
+    if (!name.toLowerCase().contains("test")) {
+        options.errorprone.disableAllChecks = true
+        options.errorprone.disableAllWarnings = true
+        options.errorprone {
+            check("NullAway", CheckSeverity.ERROR)
+            check("AnnotatorScanner", CheckSeverity.ERROR)
+            option("NullAway:AnnotatedPackages", "test")
+            option("NullAway:SerializeFixMetadata", "true")
```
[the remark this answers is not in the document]
nimakarimipour: Thank you for noting this. I will add these flags, in the followup PR for adding more unit tests.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
```diff
@@ -301,6 +307,17 @@ public Config(String[] args) {
     deactivateInference.setRequired(false);
     options.addOption(deactivateInference);
 
+    // Region detection for code generators
+    // Lombok
+    Option activateRegionDetectionForLombok =
+        new Option(
+            "ardl",
+            "activate-region-detection-lombok",
+            false,
```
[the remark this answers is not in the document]
nimakarimipour: 94f67a76c8368165cb2286e5fe3849961d2f4b7a activated it by default.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/CompoundTracker.java
```diff
@@ -28,43 +28,42 @@
...
-    MethodRegionTracker methodRegionTracker = new MethodRegionTracker(config, modules, tree);
-    this.trackers.add(new FieldRegionTracker(config, modules));
-    this.trackers.add(methodRegionTracker);
-    this.trackers.add(new ParameterRegionTracker(tree, methodRegionTracker));
+    this.trackers =
+        ImmutableSet.of(
+            new FieldRegionTracker(config, info),
+            methodRegionTracker,
+            new ParameterRegionTracker(tree, methodRegionTracker));
+    Set<GeneratedRegionTracker> generatedRegionTrackers = new HashSet<>();
```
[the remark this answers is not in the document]
nimakarimipour: d0f30c6ca12993e073fd811245d01bc260b12985

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
nimakarimipour: I looked at generated code by `@Builder` and I did not see any generated code getting reused again. But I think as discussed in the follow up PR for more tests, we can also resolve this issue if it comes up 🙂

annotator-core/src/test/java/edu/ucr/cs/riple/core/LombokTest.java
```diff
@@ -0,0 +1,61 @@
...
+            "   }",
+            "   public Object m2(){",
+            "       return getF();", // Should be error
+            "   }",
+            "}")
+        .toDepth(1)
+        .addExpectedReports(
+            new TReport(new OnField("Main.java", "test.Main", Collections.singleton("f")), 1))
+        .start();
+  }
```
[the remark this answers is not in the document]
nimakarimipour: As discussed offline, we decided to have these tests in a follow up PR.

nimakarimipour:
@lazaroclapp Thank you for the review. I resolved all your comments and this is ready for another round. Thank you 🙂.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/GeneratedRegionTracker.java
```diff
@@ -0,0 +1,33 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package edu.ucr.cs.riple.core.metadata.trackers.generatedcode;
+
+import edu.ucr.cs.riple.core.metadata.trackers.Region;
+import java.util.Set;
+
+public interface GeneratedRegionTracker {
```
msridhar, line 30: Brief Javadoc would be good

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Tracker for extending potentially impacted regions for elements which will be use in generated
+ * code by <a href="https://projectlombok.org">Lombok</a>. Lombok automatically propagates
+ * {@code @Nullable} annotation on fields to getter methods, therefore, extends the set of
+ * potentially impacted regions to all callers of that method as well. This tracker, will include
+ * all callers of any method which uses an element in a lombok generated code. This will guarantee
```
msridhar, line 42: If I understand correctly, I think this should be clearer:
```suggestion
 * all callers of any method region in lombok generated code. This will guarantee
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
```diff
@@ -0,0 +1,77 @@
...
+        .filter(region -> region.sourceType.equals(SourceType.LOMBOK) && region.isOnMethod())
+        // find the corresponding method for the region.
+        .map(region -> methodDeclarationTree.findNode(region.member, region.clazz))
+        .filter(Objects::nonNull)
+        // get method location.
+        .map(methodNode -> methodNode.location)
+        // add potentially impacted regions for the collected methods.
+        .flatMap(
+            onMethod -> {
+              Optional<Set<Region>> ans = tracker.getRegions(onMethod);
```
msridhar, line 72: Not related to this PR, but it seems strange that the `getRegions()` method returns an `Optional<Set<Region>>` rather than just a `Set<Region>` that could be empty.  Maybe fix in a follow up

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/SymbolUtil.java
```diff
@@ -154,4 +158,51 @@ private static Stream<? extends AnnotationMirror> getTypeUseAnnotations(Symbol s
...
+   * field, method or {@code null} if path leads to a static block initialization.
+   *
+   * @param path Path leading to region.
+   * @param enclosingClass Enclosing class of the path.
+   * @return Region member.
+   */
+  @Nullable
+  public static Symbol locateRegionMemberForSymbolAtPath(
+      TreePath path, Symbol.ClassSymbol enclosingClass) {
+    // Class is not generated by lombok. Let's find the region member, can be a field, method or
```
msridhar, line 173: Why is there a comment about lombok here?  Feels somewhat unrelated to what this method is doing

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/LombokGeneratedCodeDetector.java
```diff
@@ -0,0 +1,78 @@
...
+ * Detector for generated code by <a href="https://projectlombok.org">Lombok</a>. The basic
+ * assumption used here is that all generated code by Lombok is inside a method annotated with <a
+ * href="https://projectlombok.org/api/lombok/Generated">lombok.Generated</a> annotation. Please
+ * note that {@code "lombok.addLombokGeneratedAnnotation = true"} should be set in lombok
+ * configurations while building target module to add the required annotations.
+ */
+public class LombokGeneratedCodeDetector implements GeneratedCodeDetector {
+
+  /** Associated source type for Lombok in region serializations in <i>SOURCE_TYPE</i> column. */
+  private static final SourceType sourceType = SourceType.LOMBOK;
```
msridhar, line 43: I don't think we need this constant, since it's only used once.  Just return `SourceType.LOMBOK` from the method

msridhar (approved):
Only minor comments this time.  Approving assuming they will be addressed before landing

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/LombokTracker.java
nimakarimipour: @msridhar I wanted to distinguish between trackers "not having any knowledge about the passed element" and "having data but not having any actual usage". But this was required in the old designs and should be updated in follow up PRs.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/GeneratedRegionTracker.java
nimakarimipour: d97d3c4660d066e421bfa5c0580d53ddd80bee9d

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/SymbolUtil.java
nimakarimipour: Sorry thanks for noticing this. This was copied from existing method. 37387665f141cec005997acb344854d6dea6685a

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/LombokGeneratedCodeDetector.java
nimakarimipour: 71cc44387e60b52e1f588bee27b4295934846ef7

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/GeneratedRegionTracker.java
```diff
@@ -0,0 +1,46 @@
...
+ * href="https://projectlombok.org">Lombok</a>)
+ */
+public interface GeneratedRegionTracker {
+
+  /**
+   * Extents set of received potentially impacted regions with generated regions that can be
+   * impacted as well.
+   *
+   * @param regions Set of potentially impacted regions.
+   * @return Set of potentially impacted regions including all knowing potentially impacted
```
lazaroclapp, line 42:
```suggestion
   * @return Set of potentially impacted regions including all known potentially impacted
```

But also this is both hard to parse and - I think - inaccurate.

You are not extending the list exclusively with generated regions, but, rather whenever a region in the initial set (a region affected by a fix) is a generated region, we are also adding every region that would be impacted by a fix within the generated region, since the annotation processor might propagate changes from the original fix to this generated region independent of the annotator's own logic. Is this correct? Is there a clearer way to explain this? (cc: @msridhar )

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/GeneratedRegionTracker.java
```diff
@@ -0,0 +1,46 @@
...
+
+package edu.ucr.cs.riple.core.metadata.trackers.generatedcode;
+
+import edu.ucr.cs.riple.core.metadata.trackers.Region;
+import java.util.Set;
+
+/**
+ * Interface for Trackers which can extend the set of potentially impacted regions with generated
+ * regions by a processor that do not exist in source code. (e.g. <a
+ * href="https://projectlombok.org">Lombok</a>)
```
msridhar, line 33:
```suggestion
 * Interface for Trackers which can extend the set of potentially impacted regions to account for regions generated by an annotation processor that do not exist in source code. (e.g. <a
 * href="https://projectlombok.org">Lombok</a>)
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/generatedcode/GeneratedRegionTracker.java
```diff
@@ -0,0 +1,46 @@
...
+public interface GeneratedRegionTracker {
+
+  /**
+   * Extents set of received potentially impacted regions with generated regions that can be
+   * impacted as well.
+   *
+   * @param regions Set of potentially impacted regions.
+   * @return Set of potentially impacted regions including all knowing potentially impacted
+   *     generated regions by this tracker.
+   */
```
msridhar, line 44: I think this might work better (and we don't need to redundantly do `@param` and `@return`):
```suggestion
   * Given an input set of impacted regions, returns an extended set of potentially-impacted regions, accounting for the properties of impacted regions in code generated by an annotation processor.
   */
```
Needs to be formatted

msridhar (approved):

PR #120, opened 2022-12-28
Small conflict graph cleanup

msridhar:
For this code it's cleaner to use a loop than `forEach()` plus a mutable array reference.


nimakarimipour:
@msridhar Thank you very much, LGTM 🙂.

nimakarimipour (approved):

PR #121, opened 2023-01-01
Change all path types to Path from String (Resolves flaky CI issue)

nimakarimipour:
This PR converts all path types to `Path` from `String` stored in all data structures. It also uses `deserializePath()` method instead of `Paths.get()` to parse a string to a `Path` instance. This will resolve the flaky CI issue.


injector/src/main/java/edu/ucr/cs/riple/injector/location/OnMethod.java
```diff
@@ -41,12 +42,16 @@ public class OnMethod extends Location {
...
 
-  public OnMethod(String path, String clazz, String method) {
+  public OnMethod(Path path, String clazz, String method) {
     super(LocationType.METHOD, path, clazz);
     this.method = method;
     this.matcher = new SignatureMatcher(method);
   }
 
+  public OnMethod(String path, String clazz, String method) {
+    this(Helper.deserializePath(path), clazz, method);
```
lazaroclapp, line 52: Why have both this and the explicit call to `deserializePath` in `MethodDeclarationTree.java`? Maybe is better to only take a `Path` in the constructor here and force the caller to deserialize?

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnParameter.java
```diff
@@ -44,13 +46,17 @@ public class OnParameter extends Location {
...
 
-  public OnParameter(String path, String clazz, String method, int index) {
+  public OnParameter(Path path, String clazz, String method, int index) {
     super(LocationType.PARAMETER, path, clazz);
     this.method = method;
     this.index = index;
     this.matcher = new SignatureMatcher(method);
   }
 
+  public OnParameter(String path, String clazz, String method, int index) {
```
lazaroclapp, line 56: Same question as above, and similar for `OnField`, I suppose.

injector/src/main/java/edu/ucr/cs/riple/injector/Injector.java
```diff
@@ -70,8 +69,8 @@ public <T extends Change> Set<FileOffsetStore> start(Set<T> changes) {
         (path, changeList) -> {
           CompilationUnit tree;
           try {
-            tree = LexicalPreservingPrinter.setup(StaticJavaParser.parse(new File(path)));
-          } catch (FileNotFoundException exception) {
+            tree = LexicalPreservingPrinter.setup(StaticJavaParser.parse(path));
+          } catch (IOException exception) {
```
lazaroclapp, line 73: Why are we switching to a more general exception here? Is there something more precise we check for? Also, is `LexicalPreservingPrinter` still in use?

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -395,16 +395,22 @@ public static String getPackageName(String name) {
...
    *
-   * @param value Path to file.
+   * @param path Path to file.
    * @return The modified Path.
    */
-  public static String extractPath(String value) {
-    if (!new File(value).exists() && value.startsWith("file:")) {
-      return value.substring("file:".length());
+  public static Path deserializePath(String path) {
+    if (path.startsWith("file://")) {
```
lazaroclapp, line 404: We should probably also test for any other URI scheme and throw an error, no? What happens with Jar URIs here?

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -395,16 +395,22 @@ public static String getPackageName(String name) {
...
+      path = path.substring("file://".length());
     }
-    return value;
+    // Keep only one occurrence of "/" from the beginning if more than one exists.
```
lazaroclapp, line 407: Does this happen after checking for `file://`?

As I see it, we either get `file://some/path/foo.bar` or `/some/path/foo.bar`. Are we getting _any_ other formats here?

This actually looks like in some places we are serializing filesystem paths and in other's URIs (e.g. somewhere you are doing Path -> String -> [output] and in others Path -> URI -> String -> [output]), so I think there might be a deeper data serialization issue here that this method is somewhat papering over (by parsing the string representations of two different java objects, each with its own parsing and serialization mechanisms to the same thing, which seems more brittle than ensuring we are using either Paths or URIs consistently when we serialize stuff).

lazaroclapp (commented):
Did a quick pass, I am a bit confused about _why_ we are getting serialized data with these two separate formats. Might be worth looking at where this data gets serialized and check if we aren't printing `URI` objects in some places (default string conversion of `uri://path`) and `Path` objects in others...

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnMethod.java
nimakarimipour: Actually I wanted to that, but then we have to change expected output for all unit tests of `Injector` (more than 100 tests). I thought this might make this PR too large and we can do it in follow up. Please let me know if you think it is better to do it in the same PR. Thank you.

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnParameter.java
nimakarimipour: Please [here](https://github.com/ucr-riple/NullAwayAnnotator/pull/121#discussion_r1059974805).

injector/src/main/java/edu/ucr/cs/riple/injector/Injector.java
nimakarimipour: Yes, this method specifically throws a more general exception, also I just tested removing call to `LexicalPreservingPrinter` and everything is working as expected, I will remove it in a follow up PR.

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: Yes, direct URIs to strings are coming from [here](https://github.com/uber/NullAway/blob/d6d2978094acec39b50070daae75606c0a8c01a1/nullaway/src/main/java/com/uber/nullaway/fixserialization/location/FieldLocation.java#L48). We should change that in the upcoming version as discussed offline to serialize `realPath`.

Regarding covering all cases, we can either land this PR now to re-enable CI for #111, #113 and then update all path serializations in all annotator code and NullAway to serialize `realPath` in a follow up, or cover all cases within this PR. Please let me know which one you prefer. Thank you

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: Added case for jar paths: 2ae07c2f80a6756de35bd5095b15027f05c323bc

nimakarimipour:
@lazaroclapp Thank you for the review, this is ready for another round.

PR #125, opened 2023-01-03
Resolves #124 by updating class search for enum constants

### LINKED ISSUE #124: Injector misses enums constants while other anonymous classes exists
**Describe the bug**
According to java language specifics, Enum constants are first members of enums. In javaparser structure, enum constants are stored after all other members of enum which does not conform to how javac assigns flat names. We should first prioritize locating enums constants over other anonymous classes. As shown in the example below, enum constants are numbered before other existing anonymous classes, but in javaparser data structures, in the list of child nodes, enum constants are the last elements.

This will cause injector to annotate a wrong method e.g. `EC2#bar()`, instead of `Anonymous#bar()` or miss the annotation.
```java
public enum Main {

    EC1 { --------------------------------//Main$1.
        @Override
        public String bar() {
            return "EC1 IMPL";
        }
    },

    EC2 { --------------------------------//Main$2.
        @Override
        public String bar() {
            return "EC2 IMPL";
        }
    };

    A a = new A() {-----------------------//Main$3.
        @Override
        public String bar() {
            return null;
        }
    };

    public final void foo() {
        System.out.println("TEST");
    }

    public abstract String bar();
}

interface A {
    String bar();
}
```

nimakarimipour:
This PR resolves #124 by updating class search algorithm with prioritizing enum constants while doing a tree traversal. This will help injector to assign correct flat names to classes in a tree by visiting enum constants before other existing anonymous classes.

### DIFF
--- a/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
+++ b/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
@@ -224,7 +224,7 @@ public class Helper {
       throws TargetClassNotFound {
     final List<Node> candidates = new ArrayList<>();
     if (cursor instanceof EnumDeclaration) {
-      // According to java language specifics, Enum constants are first members of enums. In
+      // According to the Java language specification, enum constants are the first members of enums. In
       // javaparser structure, enum constants are stored after all other members of enum which does
       // not conform to how javac assigns flat names. We should prioritize visiting enums
       // constants first.
--- a/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
+++ b/injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
@@ -225,10 +225,9 @@ public class Helper {
     final List<Node> candidates = new ArrayList<>();
     if (cursor instanceof EnumDeclaration) {
       // According to the Java language specification, enum constants are the first members of
-      // enums. In
-      // javaparser structure, enum constants are stored after all other members of enum which does
-      // not conform to how javac assigns flat names. We should prioritize visiting enums
-      // constants first.
+      // enums. In JavaParser's data structures, enum constants are stored after all other members
+      // of the enum which does not conform to how javac assigns flat names. We prioritize visiting
+      // enum constants first.
       NodeList<EnumConstantDeclaration> constants = ((EnumDeclaration) cursor).getEntries();
       if (index < constants.size()) {
         return constants.get(index);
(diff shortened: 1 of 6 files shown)


injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -212,15 +213,27 @@ private static boolean isDeclarationWithName(Node node, String name) {
...
-   * @return anonymous class at specific index.
+   * @return anonymous class or enum constant at specific index.
    * @throws TargetClassNotFound if the target class is not found.
    */
-  private static Node findAnonymousClass(Node cursor, int index) throws TargetClassNotFound {
+  private static Node findAnonymousClassOrEnumConstant(Node cursor, int index)
+      throws TargetClassNotFound {
     final List<Node> candidates = new ArrayList<>();
+    if (cursor instanceof EnumDeclaration) {
+      // According to java language specifics, Enum constants are first members of enums. In
```
msridhar, line 227:
```suggestion
      // According to the Java language specification, enum constants are the first members of enums. In
```

msridhar (approved):
One minor tweak to a comment, but otherwise this LGTM!

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -212,15 +213,28 @@ private static boolean isDeclarationWithName(Node node, String name) {
...
+   * @return anonymous class or enum constant at specific index.
    * @throws TargetClassNotFound if the target class is not found.
    */
-  private static Node findAnonymousClass(Node cursor, int index) throws TargetClassNotFound {
+  private static Node findAnonymousClassOrEnumConstant(Node cursor, int index)
+      throws TargetClassNotFound {
     final List<Node> candidates = new ArrayList<>();
+    if (cursor instanceof EnumDeclaration) {
+      // According to the Java language specification, enum constants are the first members of
+      // enums. In
```
lazaroclapp, line 228: Might need some manual adjusting here to properly reformat this comment.

nimakarimipour:
@lazaroclapp Please let me know if I should land this, thank you very much.

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
msridhar: Fixed in https://github.com/ucr-riple/NullAwayAnnotator/pull/125/commits/ed14b51e7fd9f5d0080de1ba9167f6356e9d8b47

msridhar:
@nimakarimipour you can go ahead and land this now, I addressed @lazaroclapp's comment

PR #127, opened 2023-01-04
Configure CI to run all tests in all configurations

nimakarimipour:
This updates CI jobs to run all tests in all three configurations.

We have configurations below for running tests:
1. `Parallel Processing`: refers to computing effectiveness of multiple fixes simultaneously using graph coloring.
2. `Caching`: refers to caching fix’s impacts and reuse in next iterations.

This PR updates CI to run all tests in three combinations below with flags above:
|                                        |     Parallel Processing Disabled     |     Parallel Processing Enabled     |
| :---------------------: | :--------------------------------:| :-------------------------------: |
|  **Caching Disabled** |                           ✓                            |                           ✓                           | 
|  **Caching Enabled**  |                           X                             |                           ✓                           |  

In the current architecture of unit tests, we have a set of unit tests (that each will be converted to a standalone gradle project) which the internal state of Annotator is verified. Currently, both parallel processing and caching is activated. This PR updates the CI by rerunning unit tests to verify the internal state of the Annotator is consistent among these configurations as they are merely optimization and should not have any effect on the impact computation.


annotator-core/src/test/java/edu/ucr/cs/riple/core/tools/CoreTestHelper.java
```diff
@@ -312,6 +316,24 @@ public void makeAnnotatorConfigFile(Path configPath) {
...
+   *
+   * @param environmentVariableName environment variable name.
+   * @return true if the environment variable is not set, or negate of the variable if set.
+   */
+  private boolean getModeFromEnvironment(String environmentVariableName) {
+    String value = System.getenv(environmentVariableName);
+    if (value == null || value.isEmpty()) {
+      return true;
+    }
+    return !Boolean.parseBoolean(value);
```
lazaroclapp, line 334: Can we move the negation the callsites? I am just thinking if we will ever have other environment variables here, we might not _always_ keep the convention that all are `ANNOTATOR_TEST_DISABLE_{FEATURE}` flags. 

And either way, this seems strictly more clear:

```
    // Both of these are enabled unless specified otherwise in env vars:
    builder.useCacheImpact = !getModeFromEnvironment("ANNOTATOR_TEST_DISABLE_CACHING");
    builder.useParallelProcessor =
        !getModeFromEnvironment("ANNOTATOR_TEST_DISABLE_PARALLEL_PROCESSING");
```

And we get to simplify the above comment.

nimakarimipour: @lazaroclapp The reason that I did the negation inside the function was to get rid of the warnings from IDE that the return value is always inverted 🙂! But you are absolutely right, and for this case it does not make sense to write this method in this way especially about the convention `ANNOTATOR_TEST_DISABLE_{FEATURE} `. Moved the negation to call sites c3c8a2f8105bffe29dac8045a73c33fde0e835b7

nimakarimipour:
@lazaroclapp Thank you for the review, this is ready for another round.

lazaroclapp (approved):
LGTM! 🚀

nimakarimipour:
@lazaroclapp Thank you very much for the review, will land it now 🙂.

PR #132, opened 2023-02-09
Extend impacted regions by a method change

nimakarimipour:
This pull request extends the set of impacted regions for changes made to methods. Prior to this pull request, only call sites (and overriding methods) were taken into consideration as impacted regions for changes to methods. However, the use of lambda expressions or method references can actually extend the set of impacted regions and these should also be included. See example below:

```java
public class Main {

    public void f(Foo r){ }

    public void baz1(){
        f(new Foo(){
            public void bar(Object o){
                t(o);
            }
        });
    }

    public void baz2(){
        f(e -> t(e));
    }

     public void baz3(){
        f(this::t);
    }

    public void t(Object o){ }
}

interface Foo{
    void bar(Object o);
}
```

For example, if the `o` parameter of the `bar` method in `Foo` is annotated as `@Nullable`, the regions `baz2` and `baz3` were previously not included as impacted regions since there is no trace of bar in these methods in the current implementation. As a result, the triggered errors (such as `passing @Nullable for t(Object)`) will not be caught by the annotator. This pull request addresses this issue and extends the set of impacted regions to include `baz2()` and `baz3()` for changes made to `Foo#bar()`.

This PR also renames the output `call_graph.tsv` to `method_impacted_region_map.tsv` to better reflect its new content as it is no longer limited to just call sites, but now also includes the additional regions.

### DIFF
--- a/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
+++ b/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
@@ -453,7 +453,7 @@ public class CoreTest extends BaseCoreTest {
   }
 
   @Test
-  public void impactedLambdaAndMemberReferenceTest() {
+  public void impactedLambdaAndMemberReferenceParameterNullableTest() {
     coreTestHelper
         .addInputLines(
             "Main.java",
@@ -475,7 +475,7 @@ public class CoreTest extends BaseCoreTest {
             "   public void baz3() {",
             "       f(this::take);",
             "   }",
-            "   public Object take(Object o){ return null; }",
+            "   public void take(Object o){ }",
             "   public void passNull(Foo foo) {",
             "       foo.bar(null);",
             "   }",
@@ -483,8 +483,34 @@ public class CoreTest extends BaseCoreTest {
         .addInputLines(
             "Foo.java", "package test;", "public interface Foo{", "     void bar(Object o);", "}")
         .addExpectedReports(
-            new TReport(new OnParameter("Foo.java", "test.Foo", "bar(java.lang.Object)", 0), 2),
-            new TReport(new OnMethod("Main.java", "test.Main", "take(java.lang.Object)"), 0))
+            new TReport(new OnParameter("Foo.java", "test.Foo", "bar(java.lang.Object)", 0), 2))
+        .toDepth(1)
+        .start();
+  }
+
+  @Test
+  public void impactedLambdaAndMemberReferenceReturnNullableTest() {
+    coreTestHelper
+        .addInputLines(
+            "Main.java",
+            "package test;",
+            "import javax.annotation.Nullable;",
+            "public class Main {",
+            "   public void f(Foo r){ }",
+            "   public void baz1() {",
+            "       f(e -> bar(e));",
+            "   }",
+            "   public void baz2() {",
+            "       f(this::bar);",
+            "   }",
+            "   public Object bar(Object o){",
+            "       return null;",
+            "   }",
+            "}")
+        .addInputLines(
+            "Foo.java", "package test;", "public interface Foo{", "     Object m(Object o);", "}")
+        .addExpectedReports(
+            new TReport(new OnMethod("Foo.java", "test.Main", "bar(java.lang.Object)"), 1))
         .toDepth(1)
         .start();
   }
--- a/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
+++ b/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
@@ -436,4 +436,40 @@ public class CoreTest extends BaseCoreTest {
         .enableForceResolve()
         .start();
   }
+
+  @Test
+  public void impactedLambdaAndMemberReferenceTest() {
+    coreTestHelper
+        .addInputLines(
+            "Main.java",
+            "package test;",
+            "import javax.annotation.Nullable;",
+            "public class Main {",
+            "   public void f(Foo r){ }",
+            "   public void baz1(){",
+            "     f(new Foo(){",
+            "       @Override",
+            "       public void bar(Object o){",
+            "         take(o);",
+            "       }",
+            "     });",
+            "   }",
+            "   public void baz2() {",
+            "       f(e -> take(e));",
+            "   }",
+            "   public void baz3() {",
+            "       f(this::take);",
+            "   }",
+            "   public void take(Object o){}",
+            "   public void passNull(Foo foo) {",
+            "       foo.bar(null);",
+            "   }",
+            "}")
+        .addInputLines(
+            "Foo.java", "package test;", "public interface Foo{", "     void bar(Object o);", "}")
+        .addExpectedReports(
+            new TReport(new OnParameter("Foo.java", "test.Foo", "bar(java.lang.Object)", 0), 2))
+        .toDepth(1)
+        .start();
+  }
 }
--- a/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
+++ b/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
@@ -460,7 +460,7 @@ public class CoreTest extends BaseCoreTest {
             "   public void baz3() {",
             "       f(this::take);",
             "   }",
-            "   public void take(Object o){}",
+            "   public Object take(Object o){ return null; }",
             "   public void passNull(Foo foo) {",
             "       foo.bar(null);",
             "   }",
@@ -468,7 +468,8 @@ public class CoreTest extends BaseCoreTest {
         .addInputLines(
             "Foo.java", "package test;", "public interface Foo{", "     void bar(Object o);", "}")
         .addExpectedReports(
-            new TReport(new OnParameter("Foo.java", "test.Foo", "bar(java.lang.Object)", 0), 2))
+            new TReport(new OnParameter("Foo.java", "test.Foo", "bar(java.lang.Object)", 0), 2),
+            new TReport(new OnMethod("Main.java", "test.Main", "take(java.lang.Object)"), 0))
         .toDepth(1)
         .start();
   }
--- a/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
+++ b/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
@@ -107,7 +107,7 @@ public class AnnotatorScanner extends BugChecker
     }
     config
         .getSerializer()
-        .serializeCallGraphNode(
+        .serializeImpactedRegionByMethod(
             new TrackerNode(config, ASTHelpers.getSymbol(tree), state.getPath()));
     return Description.NO_MATCH;
   }
@@ -117,7 +117,7 @@ public class AnnotatorScanner extends BugChecker
     Config config = context.getConfig();
     config
         .getSerializer()
-        .serializeCallGraphNode(
+        .serializeImpactedRegionByMethod(
             new TrackerNode(config, ASTHelpers.getSymbol(tree), state.getPath()));
     return Description.NO_MATCH;
   }
@@ -194,7 +194,8 @@ public class AnnotatorScanner extends BugChecker
         context
             .getConfig()
             .getSerializer()
-            .serializeCallGraphNode(new TrackerNode(config, calledMethod, visitorState.getPath()));
+            .serializeImpactedRegionByMethod(
+                new TrackerNode(config, calledMethod, visitorState.getPath()));
       }
     }
     return Description.NO_MATCH;
@@ -229,6 +230,6 @@ public class AnnotatorScanner extends BugChecker
     }
     config
         .getSerializer()
-        .serializeCallGraphNode(new TrackerNode(config, methodSym, state.getPath()));
+        .serializeImpactedRegionByMethod(new TrackerNode(config, methodSym, state.getPath()));
   }
 }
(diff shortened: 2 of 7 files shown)


msridhar:
Good catch.  I think the essence of the issue here is that lambdas (and method references) implicitly create an overriding method, which could have new errors when annotations are added to the functional interface method.  To account for this conservatively, we make the containing method of the lambda an impacted region.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
```diff
@@ -151,6 +169,45 @@ public Description matchMemberSelect(MemberSelectTree tree, VisitorState state)
...
+  @Override
+  public Description matchLambdaExpression(
+      LambdaExpressionTree lambdaExpressionTree, VisitorState visitorState) {
+    Config config = context.getConfig();
+    if (!config.callTrackerIsActive()) {
+      return Description.NO_MATCH;
+    }
+    // for e -> Foo.bar(e), assume that method "baz()" has been overridden. We need to serialize the
+    // region (leaf of path in visitor state) for "baz()".
+    serializeImpactedRegionForFunctionalInterface(config, lambdaExpressionTree, visitorState);
```
msridhar, line 181: Ah, ok, so the serialized region is the lambda / method reference itself?  Does this "overlap" with the region for the containing method, or is that a separate region?

msridhar: Also, is the reason you don't serialize the called method `bar` here (like you do below for a method reference) because that will get handled when visiting the body of the lambda?

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
```diff
@@ -165,4 +222,23 @@ private void serializeSymIfField(Symbol symbol, VisitorState state) {
...
+   *
+   * @param tree Given tree.
+   * @param state Visitor State.
+   */
+  private static void serializeImpactedRegionForFunctionalInterface(
+      Config config, ExpressionTree tree, VisitorState state) {
+    Symbol.MethodSymbol methodSym = SymbolUtil.getFunctionalInterfaceMethod(tree, state.getTypes());
+    if (methodSym == null) {
+      return;
+    }
```
msridhar, line 239: I think the method symbol should always be non-null here.  Should we throw an exception if we have `null` rather than just returning?

msridhar (commented):
The change mostly LGTM, but I have a couple of questions

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
nimakarimipour: @msridhar The serialized region is the region for the containing method. And yes exactly, the called method gets handled when visiting the body of lambda.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
nimakarimipour: @msridhar I am not sure about throwing an exception in Scanner. Throwing an exception here will terminate the serialization, but Annotator cannot distinguish if the serialized output is complete or incomplete (terminated by an exception) and will work on whatever serialized data which leads to bad outcomes by Annotator. The only way I can think of is to pars standard error while building the target and monitor it to see if any exception is thrown by Scanner. Please let me know if you think I should implement this interaction between Annotator and Scanner.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
msridhar: Ok I get it now.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
```diff
@@ -151,6 +169,45 @@ public Description matchMemberSelect(MemberSelectTree tree, VisitorState state)
...
+  public Description matchMemberReference(
+      MemberReferenceTree memberReferenceTree, VisitorState visitorState) {
+    Config config = context.getConfig();
+    if (!config.callTrackerIsActive()) {
+      return Description.NO_MATCH;
+    }
+    // for Foo::bar, which is shorthand for e -> Foo.bar(e), assume that method "baz()" has been
+    // overridden. We need to serialize the region (leaf of path in visitor state)
+    // for both "baz()" and the called method "bar()" as an impacted region".
+    // serialize the overridden method: "baz()"
```
msridhar, line 195:
```suggestion
    // for Foo::bar, which is shorthand for e -> Foo.bar(e), assume that method "baz()" has been
    // overridden. We need to serialize the impacted region (leaf of path in visitor state)
    // for both "baz()" and also the called method "bar()".
    // serialize the overridden method: "baz()"
```

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
msridhar: ok, it would be nice if we could at least print or log something, but I'm ok with leaving this as just a return I guess

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/Serializer.java
```diff
@@ -61,19 +61,21 @@ public class Serializer {
...
+   * Appends the string representation of the {@link TrackerNode} corresponding to method impacted
+   * region.
    *
-   * @param callGraphNode TrackerNode instance.
+   * @param methodImpactedRegion TrackerNode instance which is an impacted region for a method
+   *     change.
    */
-  public void serializeCallGraphNode(TrackerNode callGraphNode) {
-    appendToFile(callGraphNode.toString(), this.callGraphPath);
+  public void serializeImpactedRegionByMethod(TrackerNode methodImpactedRegion) {
```
msridhar, line 77: What do you think of this?
```suggestion
  public void serializeImpactedMethodRegion(TrackerNode impactedMethodRegion) {
```
To me this is clearer.

msridhar (commented):
Some comments, and see also #135

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
nimakarimipour: Sure, makes sense. f5792cee75e1c2b804341623668bb8e66f3a77e3

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/Serializer.java
nimakarimipour: @msridhar In this method, we only serialize a region that can be impacted for a change on a __method__ (not a change on a field) and the serialized region can be a method or a field. Do you still think this is clearer considering its purpose. I think `serializeImpactedMethodRegion` might sound as a method for serializing method regions that can be impacted either by change on method or field. Either way, I will update javadoc to describe this clearer once the name is finalized.

msridhar: Ah ok.  I misunderstood.  Then how about `serializeImpactedRegionForMethod`, and also add corresponding javadoc to explain?

nimakarimipour: d6a6944f4dbdbb746fe6f639e8103cad21d9dfbc

msridhar:
@nimakarimipour is this ready for another review?

nimakarimipour:
@msridhar Yes it is now.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
```diff
@@ -151,6 +169,46 @@ public Description matchMemberSelect(MemberSelectTree tree, VisitorState state)
...
+    }
+    // for Foo::bar, which is shorthand for e -> Foo.bar(e), assume that method "baz()" has been
+    // overridden. We need to serialize the impacted region (leaf of path in visitor state)
+    // for both "baz()" and also the called method "bar()".
+    // serialize the overridden method: "baz()"
+    serializeImpactedRegionForFunctionalInterface(config, memberReferenceTree, visitorState);
+    if (memberReferenceTree instanceof JCTree.JCMemberReference) {
+      Symbol calledMethod = ((JCTree.JCMemberReference) memberReferenceTree).sym;
+      if (calledMethod instanceof Symbol.MethodSymbol) {
+        // serialize the called method: "bar()"
```
msridhar, line 201: Sorry, one more question here.  Can you give an example of what goes wrong if we _don't_ serialize the called method `bar()` here?  I thought I understood it, but thinking more I'm not so sure.

annotator-scanner/src/test/java/edu/ucr/cs/riple/scanner/MethodImpactedRegionTest.java
```diff
@@ -0,0 +1,213 @@
...
+            "   Object foo() { return null; };",
+            "}")
+        .setExpectedOutputs(
+            new TrackerNodeDisplay("edu.ucr.A", "bar()", "edu.ucr.Other", "foo()"),
+            new TrackerNodeDisplay("edu.ucr.A", "bar()", "edu.ucr.Other", "Other()"))
+        .doTest();
+  }
+
+  @Test
+  public void fieldDeclaredRegionComputationAllCases() {
```
msridhar, line 80: Sorry, how does this test relate to extending impacted regions for a method change?

msridhar (commented):
couple more questions

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
nimakarimipour: Sure, please consider the code below:
For interface:
```java
interface Foo {
      void m(Object o);
}
```
and code:
```java
void f(Foo f){ }

void baz() {
     f(this::bar);
}
void bar(Object o){ return null; }
```

If we annotate parameter `o` in `Foo#m` as `@Nullable` we will get an error in ` f(this::bar);` since `o` in `m` is `@Nullable` and we are passing it to `bar`. Therefore the region for the member reference expression is a potentially impacted region. Please let me know if this makes sense or I should work on another approach to address this. Thank you.

annotator-scanner/src/test/java/edu/ucr/cs/riple/scanner/MethodImpactedRegionTest.java
nimakarimipour: Soory for the mistake, this unit test should have been inside previous landed PRs, will extract it from this PR and add it in a follow up.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
msridhar: The scenario you describe is related to overriding I think, and is handled by this call, right?

https://github.com/ucr-riple/NullAwayAnnotator/blob/441ec21c492ca72fa8d453518ac2fc8477dc4f24/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java#L197

You are describing why the method reference needs to be an impacted region for `Foo#m`, which I completely agree with.  I was hoping for an example showing why it _also_ should be an impacted region for `bar`.

nimakarimipour: `serializeImpactedRegionForFunctionalInterface` will only serialize the overriden method (`m` for region `baz()` in this case). But this region is also an impacted region for changes on `bar` as well. Please find the example below:
```java
@Nullable
Object bar(){
    return null;
}

void f(Foo f){}

void baz(){
    f(this::bar);
}
```

This will also trigger an error in `baz()` region since `bar()` is returning `@Nullable`, that without serializing the called method in method reference (`bar()` in this case ) annotator would have missed this error. Please let me know if I understood your question correctly, thank you.

msridhar: I get it now, thanks!  Here, adding a `@Nullable` annotation on the return type of `bar` will cause an override error inside `baz`.  So, `baz` must also be an impacted region for `bar`.  Makes sense

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -451,4 +451,41 @@ public void staticBlockLocalVariableInitializationTest() {
...
+            "       public void bar(Object o){",
+            "         take(o);",
+            "       }",
+            "     });",
+            "   }",
+            "   public void baz2() {",
+            "       f(e -> take(e));",
+            "   }",
+            "   public void baz3() {",
+            "       f(this::take);",
```
msridhar, line 476: Now that I understand better the method reference handling, can we somehow split this into two different tests for method references?  One test should check that if `@Nullable` is added to the parameter of the functional interface, the region enclosing the method reference is treated as impacted.  A second test should check that if `@Nullable` is added to the return type of the referenced method, the region enclosing the method reference is treated as impacted.  I think this line is somehow testing both simultaneously.  You may need to add a second functional interface to test each case separately.  With these separate tests, both the cases from our discussion about method references would be explicit in the tests

msridhar (commented):
One final comment and then this will be good to go

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: Sure, this is a lot more clear now. I added a separate unit test that is checking the impact of making the called method in `method references` get annotated as `@Nullable` and the other one is checking the impact of making the parameter in the overridden method in the interface as `@Nullable`.  3d182ceececa0472274fb076f549731b22ca4bdd

msridhar (approved):
Ok this one now LGTM, I think we can land it

PR #133, opened 2023-02-10
Add @NullUnmarked on constructors

nimakarimipour:
This PR fixes a bug which was causing annotator to only inject `@NullUnmarked` annotations on methods. This PR fixes this issue and enables Annotator to add `@NullUnmarked` annotation for errors enclosed by constructors.

### DIFF
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -238,7 +238,12 @@ public class Annotator {
             // find the corresponding method nodes.
             .map(
                 error -> {
-                  if (!error.isInitializationError() && error.getRegion().isOnCallable()) {
+                  if (error.getRegion().isOnCallable()
+                      &&
+                      // We suppress initialization errors reported on constructors using
+                      // @SuppressWarnings("NullAway.Init"). We add @NullUnmarked on constructors
+                      // only for errors in the body of the constructor.
+                      !error.isInitializationError()) {
                     return methodDeclarationTree.findNode(error.encMember(), error.encClass());
                   }
                   // For methods invoked in an initialization region, where the error is that
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -302,8 +302,8 @@ public class Annotator {
             .filter(
                 fix ->
                     fix.isOnField()
-                        && (fix.reasons.contains("FIELD_NO_INIT")
-                            || fix.reasons.contains("METHOD_NO_INIT")))
+                        && (fix.reasons.contains("METHOD_NO_INIT")
+                            || fix.reasons.contains("FIELD_NO_INIT")))
             // Filter nodes annotated with SuppressWarnings("NullAway")
             .filter(fix -> !fieldsWithSuppressWarnings.contains(fix.toField()))
             .map(
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -229,8 +229,6 @@ public class Annotator {
     Utility.buildTarget(config);
     Set<Error> remainingErrors =
         Utility.readErrorsFromOutputDirectory(config, config.target, fieldDeclarationStore);
-    System.out.println(
-        remainingErrors.size() + " errors remaining. Applying suppression annotations.");
     Set<Fix> remainingFixes = Utility.readFixesFromOutputDirectory(config, fieldDeclarationStore);
     // Collect all regions for NullUnmarked.
     // For all errors in regions which correspond to a method's body, we can add @NullUnmarked at
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
@@ -240,7 +240,7 @@ public class Annotator {
             // find the corresponding method nodes.
             .map(
                 error -> {
-                  if (error.getRegion().isOnCallable()) {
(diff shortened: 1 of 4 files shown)


annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -229,6 +229,8 @@ private void forceResolveRemainingErrors() {
     Utility.buildTarget(config);
     Set<Error> remainingErrors =
         Utility.readErrorsFromOutputDirectory(config, config.target, fieldDeclarationStore);
+    System.out.println(
+        remainingErrors.size() + " errors remaining. Applying suppression annotations.");
```
msridhar, line 233: Is this print supposed to be part of this PR?

msridhar (approved):
One minor comment but overall this LGTM!

lazaroclapp (commented):
Makes sense to me, but I'd expect at least one test case to go with this feature/fix. Do we have a good way to test these `@NullUnmarked` injections now?

annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: @msridhar Yes, I was planning to add all changes to error suppression in this PR, but it is already divided into two PRs. I can add this in a follow up PR.

nimakarimipour:
@msridhar @lazaroclapp I noticed a problem in a followup PR for this PR and updated this one. Would you please take a look at the changes. 

The plan for remaining error suppression is as below:
1. Add `@NullUnmarked` on the containing method where the error is reported except for initialization errors.
2. Add `SuppressWarings("NullAway.Init")` for initialization errors on the field __not__ on the constructors.
3. Add `SuppressWarings("NullAway")` for errors in the initializer expression for fields on the initialized field.
4. Add `NullUnmarked` at the class level for errors in the static initialization blocks
5. Add `NullUnmarked` at the class level if the error is still unresolved after step 1 to 4.

Prior to the changes I made recently, we were adding `@NullUnmarked` on the constructors for uninitialized fields. The latest changes also conforms to version `1.3.6-alpha-5`.

msridhar (approved):
LGTM though I agree with @lazaroclapp that we should have a test here

nimakarimipour:
@msridhar @lazaroclapp Please find the update test case here 5f25f1b66e08ea568e9e03f17a53fc79cbe39516. This unit test with this change would have been failed without changes in this PR.

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -174,6 +179,16 @@ public void fieldAssignNullableConstructorForceResolveEnabled() {
...
+    List<AddAnnotation> expectedAnnotations =
+        List.of(
+            new AddMarkerAnnotation(
+                new OnMethod("Main.java", "test.Main", "Main(java.lang.Object)"),
+                "org.jspecify.nullness.NullUnmarked"),
+            new AddMarkerAnnotation(
+                new OnMethod("Main.java", "test.Main", "Main(java.lang.Object,java.lang.Object)"),
+                "org.jspecify.nullness.NullUnmarked"));
+    Assert.assertEquals(
+        expectedAnnotations, coreTestHelper.getConfig().log.getInjectedAnnotations());
```
lazaroclapp, line 191: Ok, I think this test logic is correct. But also, why are we piggy-backing testing a new feature/behavior onto an existing test, rather than creating one specifically for this change?

nimakarimipour: @lazaroclapp Yes, recently I have been incorporating some new tests cases in the existing ones. The reason behind that is that the current CI is rather a heavy process due to the unit tests architecture. Each unit test is converted to a standalone `gradle` project and that makes each unit test somewhat heavy. We currently have 53 unit tests and even though the procedure is parallelized ( to the number of cores), in my laptop with 8 cores it takes more than 10 minutes. We are planning to have a CI that runs all tests with all optimization disabled and that should take even more. I think it worths keeping CI light and for new features that we would like to test if:
1. The code pattern is very small (in this case just a constructor)
2. The code pattern is not so different from other existing tests.
3. And the selected unit tests to be embedded is testing the similar behavior with new feature.

We may be able to avoid adding a new test case and just include it in existing ones. Please let me know your thoughts on this and if this is a good approach. Thank you very much.

lazaroclapp: Ok, I understand the infrastructure limitations now. I still think generally speaking such "combined" unit tests are a bad practice: a unit test, properly speaking, should be the smallest possible piece of code required to test a specific functionality in isolation. Given arbitrary engineering resources, I think it would be desirable to change the testing infra so that test cases are either much faster or this "combination" happens transparently from the point of view of a user running tests.

That said, that's obviously a massive undertaking (not to mention one which introduces its own complexities and potential errors while debugging tests), and this is a project maintained by one person, so it seems reasonable to combine test cases to some degree as above.

cc: @msridhar for any thoughts on this.

lazaroclapp (approved):

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: Thank you for the explanation and I agree this is not a good practice. We should definitely try to come up with a solution in future to resolve this. I have been planing to redesign test architecture (postponed to after  submission deadline).
The main issue is that there are calls to build tasks in the test module that creates a series of `.tsv` files. This is the bottleneck of testing that makes the process heavy and slow. I was thinking to mock this but this (the call to subprocess that builds the target project) will require defining a series of mocked outputs that might making a new unit tests too difficult as it happens multiple times with different data for each unit tests.

nimakarimipour:
@msridhar @lazaroclapp Thank you very much for the review. I will land this now. We can continue the discussion in #137

PR #134, opened 2023-02-11
Add @NullUnmarked on classes

nimakarimipour:
This PR enables annotator to add `@NullUnmarked` annotations to suppress errors inside static initialization blocks. 

It will also one last step to resolve any remaining errors after suppression injections by adding `@NullUnmarked` annotation on the enclosing class.


annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -263,6 +263,22 @@ private void forceResolveRemainingErrors() {
     // Update log.
     config.log.updateInjectedAnnotations(nullUnMarkedAnnotations);
 
+    // Collect @NullUnmarked annotations for classes for errors in static initialization blocks.
```
msridhar, line 266:
```suggestion
    // For errors within static initialization blocks, add a @NullUnmarked annotation on the enclosing class
```

msridhar, line 266: What about instance initialization blocks?  E.g., you could have:

```java
class Foo {

  {
    getNullable().toString();
  }

  @Nullable static Object getNullable() { return null; }
}
```

Will our logic handle this case too?

msridhar (commented):
couldn't read through the whole PR but here are a couple quick comments

annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: Yes it does 🙂. Updated a unit test to verify this 47998c01f5b1f7148818a15f711ba424ae84b3d9

injector/src/test/java/edu/ucr/cs/riple/injector/OnClassInjectionTest.java
```diff
@@ -0,0 +1,101 @@
...
+@RunWith(JUnit4.class)
+public class OnClassInjectionTest extends BaseInjectorTest {
+
+  @Test
+  public void onClassBasicTest() {
+    injectorTestHelper
+        .addInput(
+            "Super.java",
+            "package com.uber;",
+            "import edu.custom.NullUnmarked;",
```
lazaroclapp, line 42: Does the input need to have the import already or is this added automatically? Or is that outside the scope of this test harness?

injector/src/test/java/edu/ucr/cs/riple/injector/OnClassInjectionTest.java
```diff
@@ -0,0 +1,101 @@
...
+            "   }",
+            "}",
+            "interface I {",
+            "  void test();",
+            "}")
+        .addChanges(
+            new AddMarkerAnnotation(
+                new OnClass("Test.java", "edu.ucr.Test$1"), "edu.custom.NullUnmarked"))
+        .start();
+  }
```
lazaroclapp, line 100: How about named inner classes? Does the logic work in that case?

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -426,6 +427,9 @@ public void staticBlockLocalVariableInitializationTest() {
             "   }",
             "}",
             "class B {",
+            "   {",
+            "      foo().hashCode();",
+            "   }",
```
lazaroclapp, line 432: Do we want to test static initializer block support and instance initializer block in the same test? (if so, the test case might need renaming). Or is it perhaps better to have two separate test cases for ease of reading/debugging?

annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
```diff
@@ -314,5 +331,21 @@ private void forceResolveRemainingErrors() {
     injector.injectAnnotations(initializationSuppressWarningsAnnotations);
     // Update log.
     config.log.updateInjectedAnnotations(initializationSuppressWarningsAnnotations);
+    // Collect @NullUnmarked annotations on classes for any remaining error.
+    Utility.buildTarget(config);
+    remainingErrors =
+        Utility.readErrorsFromOutputDirectory(config, config.target, fieldDeclarationStore);
+    nullUnMarkedAnnotations =
+        remainingErrors.stream()
+            .filter(error -> !error.getRegion().isInAnonymousClass())
```
lazaroclapp, line 340: Not for this PR, but long term, in theory, this should try to find the containing method/initializer and add `@NullUnmarked` there, no? (For anonymous inner classes). It can get complicated with nested anonymous classes and the places they can appear in, so definitely not asking for that in this PR, just wondering if that's in the roadmap for `1.3.6` or *after* `1.3.6`.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
```diff
@@ -136,4 +140,22 @@ public OnField getLocationOnField(String clazz, String field) {
...
+   *
+   * @param clazz Enclosing class of the field.
+   * @return {@link edu.ucr.cs.riple.injector.location.OnClass} instance targeting the passed
+   *     classes flat name.
+   */
+  public OnClass getLocationOnClass(String clazz) {
+    FieldDeclarationInfo candidate =
+        findNodeWithHashHint(node -> node.clazz.equals(clazz), FieldDeclarationInfo.hash(clazz));
+    if (candidate == null) {
+      // field is on byte code.
```
lazaroclapp, line 156: So, this comment (and the name of the class), reference a `field`, but as far as I can tell this is looking for a class declaration and there is no field involved...

lazaroclapp (commented):
Some nits and questions, but overall it makes sense to me.

annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
msridhar: Maybe we should do some quick renaming in comments / code where you refer to static blocks, to call them initializer blocks (or just `INIT_BLOCK` to make it shorter)?

nimakarimipour: Changed to `INIT_BLOCK` cafef181d7e0027c93a3d7a525cf21f5bbd62344

injector/src/test/java/edu/ucr/cs/riple/injector/OnClassInjectionTest.java
nimakarimipour: Actually that was not necessary, removed it 139c19a3e1229bd8b6cdf090ea7140166e361be7. Thank you for noticing this.

injector/src/test/java/edu/ucr/cs/riple/injector/OnClassInjectionTest.java
nimakarimipour: Yes it does, please find the added test at 71b998bf8ce6df50e0fea6c2848334e340b8420b

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: I think the decision for now is to have them in one test until #137 is finalized, therefore I renamed the test name 344007b131ea98a58e54154d5db8e750f4c06983

annotator-core/src/main/java/edu/ucr/cs/riple/core/Annotator.java
nimakarimipour: Yes exactly. But I think the plan is to implement that in after `1.3.6`.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
nimakarimipour: Yes, this class has the required information which its main usage is to store information regarding fields. I have a plan for refactoring these data structures that store information regarding the structure of a code and combine them into one data structure. But the plan is to do that after the release of `1.3.6`. Please let me know if you would like me to change it. Thank you.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/Node.java
```diff
@@ -92,8 +92,7 @@ public Node(Fix root) {
    */
   public void setOrigins(ErrorStore errorStore) {
     this.origins =
-        errorStore.getRegionsForElements(
-            error -> error.isSingleFix() && error.getResolvingFixes().contains(root));
+        errorStore.getRegionsForElements(error -> error.getResolvingFixes().contains(root));
```
lazaroclapp, line 95: Question: what is this change about? I don't understand either what the `error.isSingleFix()` check was ensuring before, nor why are we removing it now.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/Region.java
```diff
@@ -124,12 +124,12 @@ public boolean isOnField() {
   }
 
   /**
-   * Checks if region targets a static block.
+   * Checks if region targets a static initialization block.
```
lazaroclapp, line 127:
```suggestion
   * Checks if region targets a static or instance initialization block.
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/Region.java
```diff
@@ -124,12 +124,12 @@ public boolean isOnField() {
...
    *
-   * @return true, if region is targeting a static block.
+   * @return true, if region is targeting a static initialization block.
```
lazaroclapp, line 129:
```suggestion
   * @return true, if region is targeting an initialization block.
```

annotator-core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java
```diff
@@ -40,8 +40,8 @@ public void publicMethodWithDownstreamDependencyEnabled() {
     coreTestHelper
         .addExpectedReports(
             // Change reduces errors on target by -4, but increases them in downstream dependency
-            // DepA by 3, DepB by 4 and DepC by 2. Hence, the total effect is: 5.
-            new TReport(new OnMethod("Foo.java", "test.target.Foo", "returnNullableBad(int)"), 5),
+            // DepA by 3, DepB by 4 and DepC by 3. Hence, the total effect is: 6.
+            new TReport(new OnMethod("Foo.java", "test.target.Foo", "returnNullableBad(int)"), 6),
```
lazaroclapp, line 44: What changed that changed the effects here and why is it part of this PR about `@NullUnmarked` classes? I thought effects were always computed before adding any suppressions/null-unmarked regions... is this related to the other change above I asked about? The `error.isSingleFix()` check? If so, should that fix be part of this PR?

lazaroclapp (commented):
So, mostly good except for a few nits. But one important thing is that I think there is an unrelated change that was added into this PR? Unless I am misunderstanding it. If it's a bug fix, it probably should be on a separate PR, unless it's related to adding `@NullUnmarked` on classes.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/graph/Node.java
nimakarimipour: Origins are set of regions that triggered root fix. The condition `isSingleFix` was ensuring the error __has__ exactly one fix which is not correct, initialization errors requires multiple fixes to get resolved, therefore I removed this condition.

nimakarimipour: This was fixed in #136 and the change of this PR now changed to branch in #136

annotator-core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java
nimakarimipour: Yes the bug fix is in #136  and the base of this PR now changed to branch in #136.

nimakarimipour:
@lazaroclapp This is ready for another round of review. Thank you very much.

lazaroclapp (approved):
LGTM!

PR #135, opened 2023-02-13
Rename TrackerNode to ImpactedRegion and document

msridhar:
I found #132 a bit hard to understand because I couldn't understand what exactly `TrackerNode` was for.  Here I renamed it to `ImpactedRegion` and added some documentation, which I think is helpful.  There should be no functional change here.


msridhar:
@nimakarimipour please take a look.  If it looks ok, maybe we can land this before #132?

nimakarimipour (approved):
@msridhar Thank you very much for the changes, will land it once CI passes.

PR #136, opened 2023-02-14
Update nullaway test version to 0.10.8

nimakarimipour:
This PR simply updates NullAway test version to `0.10.8` and fixes the revealed bug regarding initialization errors. One major update in `0.10.8` is serializing the actual constructor for initialization errors rather than just `null` which helped reveal this bug in tests.

This PR also update how we compute `origins` (initial set of set regions where a root fix is triggered). 

Previous to this PR we were not including initialization errors in a root fix origins, as they were reported in the null region, but with the new update on initialization serialization, we should include them (the condition is `singleFix` is removed).


lazaroclapp:
The code looks fine to me here, but could we add the context from [here](https://github.com/ucr-riple/NullAwayAnnotator/pull/134#discussion_r1107555266) to the description or otherwise explain/document what "the revealed bug regarding initialization errors" was in this PRs description?

nimakarimipour:
@msridhar @lazaroclapp This is ready for the review with the updated description. Thank you.

annotator-core/src/test/resources/templates/downstream-dependency-test/DepC/src/main/java/test/depc/DepC.java
```diff
@@ -29,6 +29,7 @@ public class DepC {
   Foo foo = new Foo();
   Object f;
 
+  // We get 2 errros here, assign_field_nullable and method_no_init for assigning nullable to f;
```
msridhar, line 32:
```suggestion
  // We get 2 errors here, assign_field_nullable and method_no_init for assigning nullable to f;
```

annotator-core/src/test/resources/templates/downstream-dependency-test/Target/src/main/java/test/target/Foo.java
```diff
@@ -25,9 +25,8 @@
 public class Foo {
   Object field = new Object();
 
-  // Making this method will resolve 5 errors here and will introduce 1 new error here and 8 new
+  // Making this method will resolve 5 errors here and will introduce 1 new error here and 10 new
```
msridhar, line 28: Do you mean this?
```suggestion
  // Making this method return type @Nullable will resolve 5 errors here and will introduce 1 new error here and 10 new
```

annotator-core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java
```diff
@@ -100,16 +100,17 @@ public void lowerBoundComputationTest() {
...
             // Only returnNullableGood triggers new errors in this fix chain (+1) and upper bound
             // should be 1
             new TReport(new OnMethod("Foo.java", "test.target.Foo", "returnNullableGood(int)"), 1),
             // Root fix triggers 1 error on downstream dependency and returnNullableBad is
-            // present in the fix tree and triggers 9 errors on downstream dependency, therefore the
-            // upper bound effect for the tree should be 10.
-            new TReport(new OnMethod("Foo.java", "test.target.Foo", "bar()"), 10))
+            // present in the fix tree and triggers 10 errors on downstream dependency, therefore
+            // the
+            // upper bound effect for the tree should be 11.
```
msridhar, line 112: bad formatting

msridhar (approved):
Minor formatting issues, I'm fine with landing once they are fixed

annotator-core/src/test/java/edu/ucr/cs/riple/core/DownstreamAnalysisTest.java
nimakarimipour: Fixed in f1598b7db98d7884869ab8ee7d41e5c291e98dd5

nimakarimipour:
@msridhar @lazaroclapp Thank you for the review,  I think this is ready to land.

PR #138, opened 2023-02-15
Add 1 unit tests for nested parameter case

nimakarimipour:
This PR is a followup for #111 which only adds two 1 test to ensure correct set of regions are selected as potentially impacted regions for a change on a parameter. For parameters, the call sites of the method, can be a potentially impacted region as well.

Consider scenario below:

```java
void foo(Object p){
       bar(p);
}

void bar(Object p)
```

assume that we have two fixes on foo#bar and `bar#p` and `foo#p`, while we are analyzing impact of making `foo#p` nullable, `bar#p` should not be annotated as nullable, otherwise we will miss the triggered error that p is nullable while invoking `bar`. This addition in effective region will make a region conflict for `foo#p` and `bar#p` and they will not be processed simultaneously.


annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/ParameterRegionTracker.java
```diff
@@ -62,7 +62,19 @@ public Optional<Set<Region>> getRegions(Location location) {
             .collect(Collectors.toSet());
     // Add the method the fix is targeting.
     regions.add(new Region(parameter.clazz, parameter.method));
-    // Add all call sites.
+    // Add all call sites. It will also reserve call sites to prevent callers pass @Nullable
```
lazaroclapp, line 65:
```suggestion
    // Add all call sites. It will also reserve call sites to prevent callers from passing @Nullable
```

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -554,4 +555,62 @@ public void impactedLambdaAndMemberReferenceReturnNullableTest() {
...
+            "}")
+        .toDepth(5)
+        .disableBailOut()
+        .addExpectedReports(
+            new TReport(
+                new OnMethod("A.java", "test.A", "foo()"),
+                -2,
+                Set.of(new OnField("A.java", "test.A", Collections.singleton("f"))),
+                Collections.emptySet()),
+            new TReport(new OnField("A.java", "test.A", Collections.singleton("f")), -1))
```
lazaroclapp, line 584: This one is due to missing initialization, correct?

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/ParameterRegionTracker.java
```diff
@@ -62,7 +62,19 @@ public Optional<Set<Region>> getRegions(Location location) {
...
+    // simultaneously while investigating parameters impact.
+    // See example below:
+    // void foo(Object o) {
+    //   bar(o);
+    // }
+    // void bar(Object o)
+    //
+    // We need to make sure that while investigating impact of `@Nullable` on bar#o, other callers
+    // are not passing `@Nullable` to bar#o. Since the corresponding error will not be triggered
```
lazaroclapp, line 74: To be clear, what we want to make sure is that while investigating the impact of adding `@Nullable` at `bar#o`, we aren't also simultaneously trying to get the impact of making some value that gets passed to `bar#o` `@Nullable`. Or, at least, that we are not trying to calculate the impact of those two changes as _independent_ changes, yet using the same NullAway run to count the errors.

So, isn't this just saying that the regions of a method argument include the method body itself and also the method bodies of all its callers? Since errors reported on that argument can change with changes affecting nullability at either it's caller (actual argument) or callee (formal argument).

Presumably, similar logic holds for method returns.

Is that all that this comment is saying? Or am I still missing what these tests are about?

lazaroclapp (commented):
Two questions and some nits below, since I am still not sure I get the explanation in the comment. Also, I'd suggest making the PR title a bit more specific if possible, but might need to wait until I understand the change before making any specific suggestions there.

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: Yes, but I just remembered that this test will fail for configuration caching disabled. I moved it to a separate PR which includes the fix as well. Please see #148.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/trackers/ParameterRegionTracker.java
nimakarimipour: Yes exactly, but we aren't trying to use the same NullAway run to count the errors, we are actually forcing them to investigate the impacts in two different runs of NullAway.

PR #144, opened 2023-02-17
Add missing flag control for constructor invocation serialization

nimakarimipour:
This PR adds the missing flag control for serialization for constructor invocation. It also adds a configuration to test to verify no output is created once all flags are turned off.


msridhar (approved):
Honestly, I'm not really sure why we need these flags to control what gets serialized at a fine-grained level.  @nimakarimipour is that documented somewhere?  But in any case, given that we need the flags, this PR looks fine to land

nimakarimipour:
@msridhar Yes, we don't need these control at a fine grained level (they were required for previous versions) and one of the immediate changes I would like to do is to change this and add only one flag which either enables or disables all serializations.

nimakarimipour:
@msridhar Added the issue here #150.

PR #148, opened 2023-02-22
Update impacted region detection to include constructors for uninitialized fields.

nimakarimipour:
This PR update the set of potentially impacted regions for uninitialized fields at declaration.

Following update to NullAway `0.10.8` initialization errors are reported in constructor regions rather than the `null` regions. For fields which are not initialized at the declaration, all constructors can be a potentially impacted region as if it is possible that the corresponding initialization error get resolved once the field is annotated as `@Nullable`.

Please consider example below:

```java
public class Foo {
    
     Object bar;
    
     public Foo() { } 
}
```

Since there is no trace of usage of `bar` in `Foo()`, currently we do not consider `Foo()` as a potentially impacted region for changes on `bar` whereas, if `bar` get annotated as `@Nullable`, an error will be resolved on `Foo()`. This PR updates the logic for detecting potentially impacted region for uninitialized fields at declaration. 

Please note that for initialized fields at declaration, constructors are not responsible for initializing the field, therefore, we do not get any initialization errors on constructors.

This PR updates the internal data structure for `FieldDeclarationStore` to keep a record of initialized fields at declaration and `MethodDeclarationTree` to identify constructors.

### DIFF
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
@@ -155,7 +155,8 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
    *     initialized.
    */
   public boolean isUninitializedField(OnField field) {
-    // According to javadoc, Multimap.get() returns an empty collection if key is not found, therefore,
+    // According to javadoc, Multimap.get() returns an empty collection if key is not found,
+    // therefore,
     // we do not need to check for key existence.
     return !Collections.disjoint(uninitializedFields.get(field.clazz), field.variables);
   }
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
@@ -155,7 +155,7 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
    *     initialized.
    */
   public boolean isUninitializedField(OnField field) {
-    // According to javadoc, map.get() returns an empty collection if key is not found, therefore,
+    // According to javadoc, Multimap.get() returns an empty collection if key is not found, therefore,
     // we do not need to check for key existence.
     return !Collections.disjoint(uninitializedFields.get(field.clazz), field.variables);
   }
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
@@ -44,7 +44,7 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
 
   /**
    * A map from class flat name to a set of field names that are declared in that class but not
-   * initialized.
+   * initialized at declaration.
    */
   private Map<String, Set<String>> uninitializedFields;
 
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
@@ -7,6 +7,8 @@ import com.github.javaparser.ast.body.BodyDeclaration;
 import com.github.javaparser.ast.body.VariableDeclarator;
 import com.github.javaparser.ast.nodeTypes.NodeWithSimpleName;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.MultimapBuilder;
 import com.google.common.collect.Sets;
 import edu.ucr.cs.riple.core.Config;
 import edu.ucr.cs.riple.core.ModuleInfo;
@@ -20,8 +22,6 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
@@ -46,7 +46,7 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
    * A map from class flat name to a set of field names that are declared in that class but not
    * initialized at declaration.
    */
-  private Map<String, Set<String>> uninitializedFields;
+  private Multimap<String, String> uninitializedFields;
 
   /**
    * Constructor for {@link FieldDeclarationStore}.
@@ -61,7 +61,7 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
   @Override
   protected void setup() {
     super.setup();
-    this.uninitializedFields = new HashMap<>();
+    this.uninitializedFields = MultimapBuilder.hashKeys().hashSetValues().build();
   }
 
   /**
@@ -109,9 +109,7 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
                         variableDeclarator -> {
                           String fieldName = variableDeclarator.getNameAsString();
                           if (variableDeclarator.getInitializer().isEmpty()) {
-                            uninitializedFields
-                                .computeIfAbsent(clazz, k -> Sets.newHashSet())
-                                .add(fieldName);
+                            uninitializedFields.put(clazz, fieldName);
                           }
                         });
                   }));
@@ -157,9 +155,8 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
    *     initialized.
    */
   public boolean isUninitializedField(OnField field) {
-    if (!uninitializedFields.containsKey(field.clazz)) {
-      return false;
-    }
+    // According to javadoc, map.get() returns an empty collection if key is not found, therefore,
+    // we do not need to check for key existence.
     return !Collections.disjoint(uninitializedFields.get(field.clazz), field.variables);
   }
 
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
@@ -156,8 +156,7 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
    */
   public boolean isUninitializedField(OnField field) {
     // According to javadoc, Multimap.get() returns an empty collection if key is not found,
-    // therefore,
-    // we do not need to check for key existence.
+    // therefore we do not need to check for key existence.
     return !Collections.disjoint(uninitializedFields.get(field.clazz), field.variables);
   }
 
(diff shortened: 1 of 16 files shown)


annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
```diff
@@ -40,6 +42,12 @@ public class FieldDeclarationStore extends MetaData<FieldDeclarationInfo> {
    */
   public static final String FILE_NAME = "class_info.tsv";
 
+  /**
+   * A map from class flat name to a set of field names that are declared in that class but not
+   * initialized at declaration.
+   */
+  private Map<String, Set<String>> uninitializedFields;
```
msridhar, line 49: Do we use Guava in the annotator or not?  If we do use Guava, we could use a [MultiMap](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Multimap.html) here instead of managing the `Set` values ourselves

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
```diff
@@ -46,7 +46,7 @@ public class MethodDeclarationTree extends MetaData<MethodNode> {
   private HashMap<Integer, MethodNode> nodes;
 
   /** Set of all classes flat name declared in module. */
-  private HashSet<String> classNames;
+  private HashMap<String, Set<MethodNode>> classConstructorMap;
```
msridhar, line 49: This could also be a `MultiMap` if we use Guava

msridhar (commented):
Mostly LGTM, just a minor thing

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
nimakarimipour: Thank you for noticing this. Yes, we do and we are using `MultiMap` in other places. This should have been `MultiMap` as well. 217ec2faae678af3a0efb253c19402a7938c26d2

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
nimakarimipour: This is also changed to multimap. 2bb3cf428caaa6d3c140106d774a3753bf57f48c

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
```diff
@@ -46,7 +48,9 @@ public class MethodDeclarationTree extends MetaData<MethodNode> {
   private HashMap<Integer, MethodNode> nodes;
 
   /** Set of all classes flat name declared in module. */
```
msridhar, line 50: Now we have two identical comments (here and line 52).  Should one be updated?

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
```diff
@@ -46,7 +48,9 @@ public class MethodDeclarationTree extends MetaData<MethodNode> {
...
-  private HashSet<String> classNames;
+  private Multimap<String, MethodNode> classConstructorMap;
```
msridhar, line 51: As a side comment, why is this class called `MethodDeclarationTree`?  I don't see how a tree is involved.  Maybe in a follow-up we can rename to `MethodDeclarationStore`?

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -421,6 +422,35 @@ public void initializationErrorWithMultipleConstructors() {
...
+            "   void run() {",
+            "       this.f = foo();",
+            "   }",
+            "   Object foo() {",
+            "        return null;",
+            "   }",
+            "}")
+        .toDepth(5)
+        .disableBailOut()
+        .addExpectedReports(
```
msridhar, line 444: I have a hard time understanding these expected reports.  Can we write a code comment explaining what we expect to happen here and what we are testing for?

msridhar (commented):
couple more comments

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
nimakarimipour: Sorry for that, fixed it in 4ad8230a3ca6e70aa3fe9531e7844d590ec68068

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
nimakarimipour: We changed the name once through #54 but even the new name is not an appropriate name. The tree structure is within the nodes that each node's parent is the immediate super method. After release of `1.3.6` , I am planning to remove this data structure along others such `FIeldDeclarationStore` as they are internal data structures of Annotator and can be replaced with existing ones.

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: Sure, please find it in 7608084c6de1b17b7e4b7bc7e1c8c08b9443737d

nimakarimipour:

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
```diff
@@ -123,6 +145,21 @@ public ImmutableSet<String> getInLineMultipleFieldDeclarationsOnField(
...
+   * Returns true if the given field is declared in the given class but not initialized at
+   * declaration. A field declaration can contain multiple inline field declarations. This method
+   * will return true if any of the declared fields is not initialized.
+   *
+   * @param field Location of field to check.
+   * @return True if at least on of the given declarations at the given location is not not
+   *     initialized.
+   */
+  public boolean isUninitializedField(OnField field) {
+    // According to javadoc, map.get() returns an empty collection if key is not found, therefore,
```
msridhar, line 158: Honestly I think we can delete this comment (it's standard behavior for a multi-map), but if you want to keep it:
```suggestion
    // According to javadoc, Multimap.get() returns an empty collection if key is not found, therefore,
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
```diff
@@ -45,8 +47,13 @@ public class MethodDeclarationTree extends MetaData<MethodNode> {
   /** Each method has a unique id across all methods. This hashmap, maps ids to nodes. */
   private HashMap<Integer, MethodNode> nodes;
 
+  /**
+   * A map from class flat name to a set of field names that are declared in that class but not
+   * initialized at declaration.
```
msridhar, line 52: Isn't this mapping class names to the constructors in the class?

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -421,6 +422,44 @@ public void initializationErrorWithMultipleConstructors() {
...
+            new TReport(
+                // adding @Nullable on foo() will trigger a fix on making field f @Nullable, so far,
+                // effect is -1 + 1 (triggered error on this.f = foo()) = 0.
+                new OnMethod("A.java", "test.A", "foo()"),
+                -2,
+                // adding @Nullable on f will resolve the triggered error by foo() and also resolves
+                // the initialization error on A() as well.
+                // Therefore, the combined effect is -1 + (-1) = -2. It resolves all existing
```
msridhar, line 452: What would have happened for this test case before this PR?  Would we have not even attempted the fix to make `f` `@Nullable` since we didn't have the impacted region connection with the initializer error?

msridhar (commented):
Sorry, couple more comments / questions

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodDeclarationTree.java
nimakarimipour: Sorry my mistake 70ec92849a5dd230d9fe4fc50ed604448d41ca83

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
nimakarimipour: Sure, please see the different approach Annotator would take before this PR:

1. Please assume that the potentially impacted region set is `P` = `{}`.
2. It will see the error on returning `null` in `foo()` 
3. Updates `P` to `{foo(), run()}`
4. Since `run()` is impacted, it can detect the triggered fix for `f`.
5. `f` is only used in `run()` therefore the combined set of potentially impacted region will be: `{foo(), run(), f}`

For fix tree: `foo() and f`:
Considering these regions, annotator resolved the error on `foo()` and introduced no error, the effect will be `-1`.


But in reality which this PR is all about:
For step `5`, although `f` is only used in `run()`, the constructor should also be added as an impacted region. Since there is a reported on `A()` about initialization of `f`. If `f` gets annotated as `@Nullable` that error will be resolved as well.
With this PR (step 1 to 4 is identical):
5. `f` is only used in `run()` and not initialized at declaration, therefore the combined set of potentially impacted region will be: `{foo(), run(), f, A()}`

For fix tree: `foo() and f`:
Considering these regions, annotator resolved the error on `foo()` and introduced no error, also resolved another error on `A()` the overall effect of the tree is `-2`. Which is the correct effect.

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
```diff
@@ -123,6 +145,22 @@ public ImmutableSet<String> getInLineMultipleFieldDeclarationsOnField(
...
+   * will return true if any of the declared fields is not initialized.
+   *
+   * @param field Location of field to check.
+   * @return True if at least on of the given declarations at the given location is not not
+   *     initialized.
+   */
+  public boolean isUninitializedField(OnField field) {
+    // According to javadoc, Multimap.get() returns an empty collection if key is not found,
+    // therefore,
+    // we do not need to check for key existence.
```
msridhar, line 160: formatting :-)

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
msridhar: Got it!  So the key is that the effect was computed incorrectly before.  Makes sense.

msridhar (approved):
Approved with one very minor comment

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/field/FieldDeclarationStore.java
nimakarimipour: Sorry again 😞 fixed it in 1646c50e660b7a45eeefbb492d21f9a5725a0033

PR #149, opened 2023-02-22
Update NullAway test version to 0.10.10

nimakarimipour:
~~This PR requires release of NullAway `0.10.10`.~~ NullAway `0.10.10` is released.

This PR updates NullAway test version to `0.10.10`. The main update in this version is on the serialization change. 

In the newest version, method signatures does not contain any annotation or type arguments, this PR adapts injector to work with this serialization to locate the method and also update scanner to use the same logic for serializing method signatures.

Please note that previously Injector was written to parse complicated cases such as case below:

```
foo(@edu.ucr.Custom({"arg1, "arg2}) java.util.Map<A.B.C, A.B.D>)
```

Wheres in the latest version of serialization for the method above, we receive input below:
```
foo(java.util.Map)
```

Therefore, the parsing logic, along the input for some unit tests had to be updated.

### DIFF
--- a/injector/src/test/java/edu/ucr/cs/riple/injector/ComprehensiveTest.java
+++ b/injector/src/test/java/edu/ucr/cs/riple/injector/ComprehensiveTest.java
@@ -319,44 +319,27 @@ public class ComprehensiveTest extends BaseInjectorTest {
         .addChanges(
             new RemoveAnnotation(
                 new OnMethod(
-                    "Main.java",
-                    "edu.ucr.Main",
-                    "foo6(Map<String, Object>, Object, Object, Object, Object)"),
+                    "Main.java", "edu.ucr.Main", "foo6(Map, Object, Object, Object, Object)"),
                 "javax.annotation.Nullable"),
             new RemoveAnnotation(
                 new OnParameter(
-                    "Main.java",
-                    "edu.ucr.Main",
-                    "foo6(Map<String, Object>, Object, Object, Object, Object)",
-                    0),
+                    "Main.java", "edu.ucr.Main", "foo6(Map, Object, Object, Object, Object)", 0),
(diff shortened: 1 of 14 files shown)


injector/src/test/java/edu/ucr/cs/riple/injector/ComprehensiveTest.java
```diff
@@ -319,44 +319,27 @@ public void methodWithMultipleLineDeclaration() {
         .addChanges(
             new RemoveAnnotation(
                 new OnMethod(
-                    "Main.java",
-                    "edu.ucr.Main",
-                    "foo6(Map<String, Object>, Object, Object, Object, Object)"),
+                    "Main.java", "edu.ucr.Main", "foo6(Map, Object, Object, Object, Object)"),
```
msridhar, line 322: Why do we just see `Map` and `Object` here rather than `java.util.Map` and `java.lang.Object`?

msridhar (commented):
Changes look fine to me, just one question

injector/src/test/java/edu/ucr/cs/riple/injector/ComprehensiveTest.java
nimakarimipour: @msridhar These tests are added through the merging of Injector to Annotator. Previously Injector was written as a separate standalone project and some of the tests cases are from previous repository. Injector can work with both full name and simple names if fully qualified names are not given as input.


Although annotator-core always gives injector fully qualified names, some of the tests are written with simple names as it was designed regardless of getting used only in annotator-core. Please let me know if you would like me to change this as Injector is no longer a standalone project. Thank you.

msridhar (approved):
LGTM

nimakarimipour:
@msridhar Thank you for the review, will land this now.

PR #163, opened 2023-03-22
Add local variable injection

nimakarimipour:
### This PR is build on #170 and #171 and cannot land before them.

This PR adds support for applying changes on local variables declared in methods.

To locate a local variable declaration expression node in the AST, information below must be passed to injector:
1. Flat name of the enclosing class.
2. Method signature of the __first__ enclosing method.
3. Local variable name.

With these information, injector is able to locate the variable declaration expression and apply the requested changes.

Annotations on local variables will be applied on the declaration and all its type arguments if exists:
For instance:  `Foo<Bar, Map<Bar, Bar>> baz`
Will be annotated as: `@Annot Foo<@Annot Bar, @Annot Map<@Annot Bar, @Annot Bar>> baz`


nimakarimipour:
Sine this PR is too old to rebase, I will close this and open a fresh one instead.

PR #165, opened 2023-03-23
Refactor Change#visit method in injector module

nimakarimipour:
This PR updates the method below:

```java
Modification visit(NodeWithAnnotations<?> node, Range range)
```
To
```java
<T extends NodeWithAnnotations<?> & NodeWithRange<?>> Modification visit(T node)
```

Since the value for `range` is retrievable from the class implementing `NodeWithAnnotations `, all calls to this method were as below:
```
visit(node, node.getRange)
```

With this update, all calls are changed to simply:
```
visit(node)
```


msridhar (approved):
Fancy :-)

nimakarimipour:
> Fancy :-)

@msridhar Thank you very much for the review, landed it. 🙂.

PR #170, opened 2023-04-04
Use visitor for json creation of locations

nimakarimipour:
This PR updates the logic for json creation from `Location` instances using the visitor pattern architecture added in #169


injector/src/main/java/edu/ucr/cs/riple/injector/location/LocationToJsonVisitor.java
```diff
@@ -0,0 +1,103 @@
...
+    VARIABLES,
+    METHOD,
+    KIND,
+    CLASS,
+    PATH,
+    INDEX
+  }
+
+  /** A singleton instance of this visitor. */
+  public static final LocationToJsonVisitor INSTANCE = new LocationToJsonVisitor();
```
msridhar, line 47: Why make this a singleton?  Since there is no state in the object, and allocating the object should be cheap, I think this is overkill / premature optimization.

nimakarimipour: Thank you for noticing this. I thought since there is not state, we should make it singleton. But as you said, this is cheap and not a good optimization overall. I fixed it here af081c111fe9b8c6dbb2fc22ea657027a93c35cf.

injector/src/main/java/edu/ucr/cs/riple/injector/location/LocationToJsonVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+  public enum KEYS {
+    VARIABLES,
+    METHOD,
+    KIND,
+    CLASS,
+    PATH,
+    INDEX
+  }
+
+  @SuppressWarnings("unchecked")
```
msridhar, line 43: Why do we need these unchecked suppressions?

injector/src/main/java/edu/ucr/cs/riple/injector/location/LocationToJsonVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Visits a location and returns a JSON object.
+   *
+   * @param location the location to visit
+   * @return a JSON object representing the location
+   */
+  public JSONObject visit(Location location) {
+    return location.accept(this, null);
+  }
```
msridhar, line 92: This method is unnecessary, right?  At any caller you could just call `Location.accept` directly with a new visitor?  I think that is standard and we can remove this method

msridhar (commented):
couple more minor comments

injector/src/main/java/edu/ucr/cs/riple/injector/location/LocationToJsonVisitor.java
nimakarimipour: This is required for putting object of any type in `JSON`. The `JSONObject` implements raw type of `Map` and every call to `put` is risky in view of javac. The only way to resolve this warning is to change the json dependency to `org.json` from `com.googlecode.json-simple:json-simple` since it is compiled with a very old javac (the latest release is for 2012). We have an open PR that is not landed yet as it has a low priority #123 that replaces the dependency with `org.json` and can get rid of all `@SuppressWarnings("unchecked")`.

injector/src/main/java/edu/ucr/cs/riple/injector/location/LocationToJsonVisitor.java
nimakarimipour: Yes exactly, removed it 6f245dab36816c252fff7114f50cb067363c6d97.

msridhar (approved):

PR #175, opened 2023-04-06
Add Context to Annotator data structures

nimakarimipour:
The `annotator-core` module contains two classes, `FieldRegistry` and` MethodRegistry`, that store structural information about the processing module. Many methods in this module require instances of these classes as arguments, which can lead to complex code. To simplify the architecture, this PR introduces the `Context` class that encapsulates both `FieldRegistry` and `MethodRegistry` in `ModuleInfo` instance. With this update, most method signatures will be changed to `foo(Context context)`, where each method can access the relevant information from the context.

Although this PR is quite extensive, its main objective is to introduce just the Context class with __no__ unrelated changes. If you agree with the proposed change outlined in the PR description and don't feel the need for a full review, please let me know, and I can proceed with merging the PR. Thank you.


msridhar (commented):
This change sounds good and I am not going to review in full detail.  My only high-level comment is that to me, it makes more sense to have the `Config` be part of the `Context` object rather than vice-versa.  I think `Context` is a common name for all the state you want to pass around and have accessible, whereas the `Config` sounds like a type for storing configuration parameters.  WDYT?  @lazaroclapp not sure if you have an opinion here

PR #184, opened 2023-04-20
Add support for modifications on local variables

nimakarimipour:
This PR adds support for applying changes on local variables declared in methods.

To locate a local variable declaration statement node in the AST, information below must be passed to injector:
1. Flat name of the enclosing class.
2. Method signature of the __closest__ enclosing method.
3. Local variable name.

With these information, injector is able to locate the variable declaration statement and apply the requested changes.


msridhar:
@nimakarimipour to be clear this PR shouldn't impact NullAway support in any way, correct?

nimakarimipour:
> @nimakarimipour to be clear this PR shouldn't impact NullAway support in any way, correct?

@msridhar Yes exactly, it does not in any way.

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -324,6 +329,34 @@ public static NodeList<BodyDeclaration<?>> getTypeDeclarationMembersByFlatName(
...
+  /**
+   * Locates a variable declaration expression in the tree of a {@link CallableDeclaration} with the
+   * given name.
+   *
+   * @param encMethod The enclosing method which the variable is declared in.
+   * @param varName The name of the variable.
+   * @return The variable declaration expression, or null if it is not found.
+   */
+  @Nullable
+  public static VariableDeclarationExpr locateVariableDeclarationExpr(
```
msridhar, line 341: It seems we do not yet support local variables inside initializer blocks or lambdas; is that correct?  If so we should probably document it somewhere.

injector/src/main/java/edu/ucr/cs/riple/injector/changes/ChangeVisitor.java
```diff
@@ -165,6 +167,45 @@ public Modification visitClass(
     return change.computeTextModificationOn(((BodyDeclaration<?>) optionalClass.get()));
   }
 
+  @Override
+  public Modification visitLocalVariable(
+      OnLocalVariable onLocalVariable, Pair<NodeList<BodyDeclaration<?>>, ASTChange> pair) {
+    final NodeList<BodyDeclaration<?>> members = pair.a;
+    final ASTChange change = pair.b;
+    final AtomicReference<Modification> ans = new AtomicReference<>();
+    members.forEach(
```
msridhar, line 176: I personally think this would be clear with a for loop rather than `forEach`.  Then you could just early-return once the right variable is found and the change is computed.

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnLocalVariable.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Represents a location for local variable element. This location is used to apply changes to a
+ * local variable.
+ *
+ * <p>Annotations on local variables will be applied on the declaration and all its type arguments
+ * if exists. For instance: {@code Foo<Bar, Map<Bar, Bar>> baz}
+ *
+ * <p>Will be annotated as: {@code @Annot Foo<@Annot Bar, @Annot Map<@Annot Bar, @Annot Bar>> baz}
```
msridhar, line 39: Is the logic for injecting these annotations in this class, or somewhere else?  I don't see it in this class.

injector/src/test/java/edu/ucr/cs/riple/injector/OnLocalVariableInjectionTest.java
```diff
@@ -0,0 +1,204 @@
...
+            "      String f2;",
+            "   }",
+            "}")
+        .expectOutput(
+            "package test;",
+            "import edu.ucr.UnTainted;",
+            "public class Foo {",
+            "   public void foo() {",
+            "      @UnTainted int f0;",
+            "      @UnTainted Bar<String, Integer, Baz<String, Integer>> f1;",
```
msridhar, line 56: Why doesn't the annotation appear on all the type arguments here?

msridhar (commented):
Few comments

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: Yes, documented the limitation. 8b433a6cd95a4d4709ced08167f8c5d16e7b48ca

injector/src/main/java/edu/ucr/cs/riple/injector/changes/ChangeVisitor.java
nimakarimipour: Sure, rewrote it with for loops f1ffe0808aca8e184ebb233863ecf67fd074881e

injector/src/test/java/edu/ucr/cs/riple/injector/OnLocalVariableInjectionTest.java
nimakarimipour: Sorry for that. This is left over before splitting this PR into two separate PRs. Annotation injections on type arguments has been implemented in #187. It also not limited to just local variables. Fixed the javadoc f1ffe0808aca8e184ebb233863ecf67fd074881e

injector/src/main/java/edu/ucr/cs/riple/injector/location/OnLocalVariable.java
nimakarimipour: Sorry for that. This is left over before splitting this PR into two separate PRs. Annotation injections on type arguments has been implemented in #187. It also not limited to just local variables. Fixed the javadoc f1ffe0808aca8e184ebb233863ecf67fd074881e

nimakarimipour:

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
msridhar: Do we have a test involving a local inside a lambda or initializer expression?  Just want to make sure we don't crash.

msridhar (approved):
Ok to land, though I do have one suggestion for a test

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
nimakarimipour: @msridhar Sure, please find the test here 5d14a4cbbbdbf22d649bf891a5bce3ff0a0c0b32
Please let me know if the test is sufficient.

msridhar: LGTM!

PR #187, opened 2023-05-09
Add support for type use annotation injection

nimakarimipour:
### ~~This PR is built upon #184~~ #184 is landed
----

This PR enables Annotator to inject/remove type use annotation.

On type-use annotation addition, injector inserts the annotation on the types name:
```java
java.util.List<String> list;
```
Will be:
```java
java.util.@Nullable List<String> list;
```
Also, our current implementation, will add the annotation on all type arguments on the declaration as well:
Final output is:
```java
java.util.@Nullable List<@Nullable String> list;
```

Within this PR the support for type-use annotation is added including both `addition` and `deletion`.

On deletion for expression:
```java
java.util.@Nullable Map<@Nullable String, java.lang.@Nullable String> list;
```
Should be:
```java
java.util.Map<String, java.lang.String> list;
```


injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddTypeUseMarkerAnnotation.java
```diff
@@ -0,0 +1,146 @@
...
+    if (!Helper.isAnnotatedWith(type, annotationExpr)) {
+      // Annotate the type.
+      Range range = findTypeRange(type);
+      if (range == null) {
+        return null;
+      }
+      modifications.add(new Insertion(annotationExpr.toString(), range.begin));
+    }
+    // Annotate type arguments. Note: We currently do not annotate Map<K, V>[] as @Annot Map<@Annot
+    // K, @Annot V>[] and will leave it as @Annot Map<K, V>[]
```
msridhar, line 80: Why do we have this special case?

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddTypeUseMarkerAnnotation.java
```diff
@@ -0,0 +1,146 @@
...
+      return false;
+    }
+    AddTypeUseMarkerAnnotation otherAdd = (AddTypeUseMarkerAnnotation) other;
+    return this.location.equals(otherAdd.location)
+        && this.annotationName.equals(otherAdd.annotationName);
+  }
+
+  /**
+   * Find the range of the given type in source code. This is used to insert the annotation before
+   * the type.
```
msridhar, line 121: What do you mean by "the type" here?  Like for `java.util.List` do you mean the position of `List`?  So maybe you mean the "simple name" in a fully-qualified type name?

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveTypeUseMarkerAnnotation.java
```diff
@@ -0,0 +1,100 @@
...
+        if (expr.equals(annotationExpr)) {
+          Optional<Range> annotRange = expr.getRange();
+          annotRange
+              .map(value -> new Deletion(expr.toString(), value.begin, value.end))
+              .ifPresent(modifications::add);
+        }
+      }
+    }
+
+    // Remove annotation from type arguments if exists e.g. List<@Annot String>
```
msridhar, line 80: Is it possible to write something like `java.util.List<java.util.Map>` where the type argument is also fully qualified?  If so, do we insert and delete annotations from the right place for the type argument?  (should have a test for this case)

injector/src/test/java/edu/ucr/cs/riple/injector/TypeUseAnnotationTest.java
```diff
@@ -0,0 +1,244 @@
...
+            "      String[] f2;",
+            "   }",
+            "}")
+        .expectOutput(
+            "package test;",
+            "import edu.ucr.UnTainted;",
+            "public class Foo<T> {",
+            "   public void foo() {",
+            "      java.util.@UnTainted Map<java.lang.@UnTainted String, @UnTainted String[]> f0;",
+            "      @UnTainted T[] f1;",
```
msridhar, line 205: So right now we do not support injecting `@UnTainted T @UnTainted []`; we only support annotating the elements of the array (the second annotation applies to the array pointer itself).  Is this correct?

msridhar (commented):
Looks good, I have a few comments

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddTypeUseMarkerAnnotation.java
nimakarimipour: Yes exactly, I mean simple name. Updated it. 1f73f49c84f56f40e96c00e6b90485352f8cf9d0

injector/src/main/java/edu/ucr/cs/riple/injector/changes/RemoveTypeUseMarkerAnnotation.java
nimakarimipour: Yes, we do have support for that, and have a test for deletion. For addition, I updated a test here: 24540c012d516b38bb30614fca77adb12ad0cdbb

injector/src/main/java/edu/ucr/cs/riple/injector/changes/AddTypeUseMarkerAnnotation.java
nimakarimipour: I thought maybe we can add this support once we see the use-case. But I think it worths to just add the support here anyway since it is very straightforward. f1b86c9e6bae5ea1c41a7c1a4700f47c9490eb6b
Also updated the test case.

injector/src/test/java/edu/ucr/cs/riple/injector/TypeUseAnnotationTest.java
nimakarimipour: Yes, we do not add the second annotation. Please let me know if you would like to add that as well.

nimakarimipour:

injector/src/test/java/edu/ucr/cs/riple/injector/TypeUseAnnotationTest.java
msridhar: At the least we need to document what we do. If we want to inject nullability type use annotations, we will need support for both. Even for tainting, if we ever have a case where the array itself may be tainted (not its contents) we will need to support both locations. But up to you on whether to do it in this PR

msridhar (approved):
One more comment, but I am also ok with landing and addressing the issues with array types in a follow-up

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
```diff
@@ -358,6 +366,58 @@ public static VariableDeclarationExpr locateVariableDeclarationExpr(
...
+      }
+    }
+    if (node instanceof Parameter) {
+      return ((Parameter) node).getType();
+    }
+    if (node instanceof VariableDeclarator) {
+      return ((VariableDeclarator) node).getType();
+    }
+    if (node instanceof ArrayType) {
+      return ((ArrayType) node).getComponentType();
```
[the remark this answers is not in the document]
nimakarimipour: Sure, this is unusual. Fixed it in this PR 7d0b20135537a0f6c995b10024d3fe5dd546fea7

injector/src/test/java/edu/ucr/cs/riple/injector/TypeUseAnnotationTest.java
nimakarimipour: I thought about this, and I think maybe we should update this in a followup PR, as it may have changes in type declaration annotations as well. I think for now, even if we instantiate a `AddMarkerAnnotation` and not a `AddTypeUseMarkerAnnotation` annotation, for arrays we annotate as below:
```java
@Annot T[]
```

Which I think `AddMarkerAnnotation` should annotate as ```T @Annot []```
Added a comment / todo to update.  669b806be6f31ddcfdc584c666bfd014e1d9c131

injector/src/main/java/edu/ucr/cs/riple/injector/Helper.java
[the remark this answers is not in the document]
nimakarimipour: Sure, Added a comment 669b806be6f31ddcfdc584c666bfd014e1d9c131

PR #193, opened 2023-05-17
Generalize Annotator to work with any checker as oracle

nimakarimipour:
This PR abstracts all tasks executed in `core` logic related to NullAway specifically and extracts them into interfaces. With this change, we can add support for a new checker (e.g. taint or other nullness checkers) by just implementing the `Checker` interface.

Below is a quick summary of tasks related specifically to NullAway:
- Backward compatibility check with different NullAway serialization versions.
- Preparation of NullAway config file before compiling a module
- Deserialization of errors reported by NullAway.
- Error suppression logic after the inference is executed.
- Preprocess step which injects `@Initializer` annotations
- Error `equals` and `hashcode` logic which is specific to NullAway (the edge case for initialization errors)


annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
```diff
@@ -46,9 +46,9 @@ public class Error {
   /** Error message. */
   public final String message;
   /** The fixes which can resolve this error (possibly empty). */
-  private final ImmutableSet<Fix> resolvingFixes;
+  protected final ImmutableSet<Fix> resolvingFixes;
```
msridhar, line 49: Just curious, should the `Error` class be abstract now?  Like do we expect each checker to have its own concrete error class?

msridhar (approved):
LGTM, just one non-blocking question.  I assume there are no behavior changes intended here, just refactoring

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/index/Error.java
nimakarimipour: At first I thought a checker can just use the default `Error` class, but it would make more sense to expect each checker have their own concrete error class (even if it is just extending the base class). Made it abstract. 6c1a5dc739f288e54754df3f4fc50ca7beb93e8a

PR #196, opened 2023-05-19
Update Scanner to serialize methods annotations on return type

nimakarimipour:
This method update Scanner to serialize all methods annotations on return type. (Will be used in the upcoming PR (#201) to enhance Lombok support).

Previous to this PR `MethodRecords` contained a flag denoting if the method is marked with `@Nullable`, with this new implementation, that flag is removed and replaced with full set of annotations.

Method's annotations on return type are serialized on their corresponding column in `method_records.tsv` file. Only annotations types are serialized and are separated by comma.

Please see example below:

```java
class Foo {
@SuppressWarnings({"A", "B"})
@Override
@lombok.Genereted
Object getBar() { ... }
```

Scanner will serialize:
```
java.lang.SuppressWarnings,java.lang.Override,lombok.Generated
```


annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRecord.java
```diff
@@ -40,8 +41,8 @@ public class MethodRecord {
   public Integer id;
   /** Location of the containing method. */
   public OnMethod location;
-  /** Is true if the method is already annotated with {@code Nullable} annotation. */
-  public boolean hasNullableAnnotation;
+  /** Set of annotations on method return type */
```
msridhar, line 44: Will this include both declaration annotations on the method and type use annotations on the return type?  Even if we don't support type use annotations right now it'd be good to know if we want all annotations to be in the same list or not.  Note that syntactically I think such annotations can be interleaved and appear in any order (and also interleaved with qualifiers like `public`)

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRegistry.java
```diff
@@ -94,7 +96,7 @@ protected Builder<MethodRecord> getBuilder() {
       node.fillInformation(
           new OnMethod(Helper.deserializePath(values[8]), values[1], values[2]),
           parentId,
-          Boolean.parseBoolean(values[5]),
+          values[5].split(ANNOTATION_DELIMITER),
```
msridhar, line 99: Presumably we need to bump the serialization version?  Do we need to worry about backwards compat?

msridhar (commented):
Couple high-level annotations

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRegistry.java
msridhar: Sorry, I realize this comment is invalid as we are only changing the Scanner output, not NullAway's output.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodRecord.java
```diff
@@ -53,15 +55,20 @@ public class MethodRecord {
   private URI uri;
   /** Unique id assigned to this method across all visited methods. */
   private final int id;
+  /**
+   * Set of annotations on the method return type. If the method has no annotations, then this field
+   * will be empty and not {@code }null}.
```
lazaroclapp, line 60:
```suggestion
   * will be empty and not {@code null}.
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRegistry.java
lazaroclapp: I believe we don't promise compatibility between anything but the exact same version of `annotator-scanner` and `annotator-core` 😉

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRecord.java
lazaroclapp: I think we need to support type use annotations at least for `@Nullable`, no? Otherwise we will have trouble when we move to JSpecify, so I'd say we want to make sure both cases are captured. Maybe worth a test with Jspecify's version of `@Nullable`?

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRecord.java
```diff
@@ -40,8 +41,8 @@ public class MethodRecord {
...
+  public String[] annotations;
```
lazaroclapp, line 45: Why is this an array if the original thing that gets serialized is a set? Likely not a big deal, but I was wondering if there was a reason not to use `ImmutableSet` here too.

nimakarimipour: That was just for simplicity since annotations are deserialized from a `tsv` column and were separated by calling `.split()` that returns an array. But it still worths making it immutable. dc785139139b843181938dab455e307fc8bcba1d

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/method/MethodRecord.java
msridhar: I looked over the relevant `SymbolUtil` method and I think this code will capture all annotations (both declaration and type use).  So I think we are ok for now.  When _injecting_ a type use annotation on a return type, we should try to put the annotation next to the return type if we can.  But that's separate from the issue being solved here.

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/out/MethodRecord.java
```diff
@@ -53,15 +55,20 @@ public class MethodRecord {
   private URI uri;
   /** Unique id assigned to this method across all visited methods. */
   private final int id;
+  /**
+   * Set of annotations on the method return type. If the method has no annotations, then this field
```
msridhar, line 59: I think technically these are annotations either on the method return type or on the method itself (in the case of a declaration annotation)?  Probably best to fix the comment?

msridhar (approved):
One comment on fixing a comment, but I don't need to re-review after that is addressed

lazaroclapp (approved):
LGTM modulo Manu's requested comment clarification!

PR #201, opened 2023-05-23
Enhance Lombok support by propagating @Nullable to corresponding getter method

### LINKED ISSUE #198: Annotator is unaware of annotations copied by Lombok in Strict Mode
**Describe the bug**
In strict mode, Annotator should not make any changes on the target module, that may introduce new errors on downstream dependencies. Although Annotator is aware of possible errors in downstream methods for a generated getter by Lombok (`getFoo()`), but it is unaware of copying mechanism by Lombok that annotations on `foo` will get copied on `getFoo` as well. Hence, it might introduce an error in downstream dependencies.

**Expected behavior**
Annotator should be aware of annotations getting copied on generated `getter` methods for fields by `Lombok`. And reject a tree, if any of their getter is creating new errors in downstream dependencies.

**Stack trace**
Not-Applicable.

 - Annotator Version [1.3.6]

**Additional context**
Not-Applicable.

nimakarimipour:
#### This PR is built upon #196 
------

This PR resolves #198 by enhancing lombok support. With this PR Annotator is aware of `@Nullable` propagation on a field to the corresponding `getter` method.

It updates `AnnotationProcessorHandlers` to extend a fix tree to account for generated code by processors in addition to extending set of impacted regions.


annotator-core/src/test/java/edu/ucr/cs/riple/core/LombokTest.java
```diff
@@ -58,4 +61,45 @@ public void regionComputationOnFieldParallelProcessingEnables() {
...
+            "       return new Main().getF();", // Should be error
+            "   }",
+            "}")
+        .withExpectedReports(
+            new TReport(
+                new OnField("Main.java", "test.Main", Collections.singleton("f")),
+                0,
+                // The copied annotation on the getF creates an unresolvable error in downstream
+                // dependency, hence, the tree should be rejected.
+                Report.Tag.REJECT))
```
msridhar, line 92: 👍

msridhar (approved):
This LGTM!  @lazaroclapp do you want to look as well?

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/region/generatedcode/LombokHandler.java
```diff
@@ -78,4 +83,65 @@ public Set<Region> extendForGeneratedRegions(Set<Region> regions) {
...
+                                  + Character.toUpperCase(name.charAt(0))
+                                  + name.substring(1)
+                                  + "()";
+                          // Check if method is lombok generated.
+                          MethodRecord getterMethod =
+                              moduleInfo
+                                  .getMethodRegistry()
+                                  .findMethodByName(onField.clazz, getterSignature);
+                          if (getterMethod == null) {
+                            // Unlikely to happen, but just in case.
```
lazaroclapp, line 110: Why is this unlikely to happen? Shouldn't it be the case for most fields of most classes unless the class or the field is explicitly annotated with one of the annotations that generate getters or the developer has manually added a getter method? Like, we'd expect most classes without Lombok, with normal private fields, to hit this case, no?

lazaroclapp (commented):
Makes sense to me! Except for one comment that I don't think is correct...

annotator-core/src/main/java/edu/ucr/cs/riple/core/metadata/region/generatedcode/LombokHandler.java
nimakarimipour: Sure, you are absolutely right. I was thinking about if the method is generated, it should still be present in the registry and is unlikely to be null. Updated the comment. aef1c184cdea0df22cd7dbab89ab50f1d9137ad8

lazaroclapp (approved):

PR #206, opened 2023-07-24
Collect std out using Redirect.INHERIT API

### LINKED ISSUE #205: Collect and print stdout output from command that builds downstream dependencies
This is the code that runs the command for building downstream dependencies:

https://github.com/ucr-riple/NullAwayAnnotator/blob/6740c7db40199fb7a57b831934f1d66629015af9/annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java#L70-L85

It collects the stderr output from the command, but not the stdout.  It would be good to have a way to collect and print stdout as well, for diagnosing issues.  Also, the current logic looks a bit weird.  There is a `while` loop that spins collecting the stderr output.  I think this loop should run until the process terminates (otherwise you may miss some output).  But if it runs until the process terminates, then I think the `p.waitFor()` call is unnecessary?  We may just want to switch to code like this:

https://stackoverflow.com/a/12200361/1126796

With Redirect.INHERIT we will get the printing of stdout / stderr for free.  (Or possibly we want to redirect stderr of the child process to stdout, and then just send everything to our stdout.)

nimakarimipour:
Resolves #205. This PR refactors std error output collection by using `Redirect.INHERIT` API. It also adds collecting std out of the child process (module builds) and prints them along the std error if `-rboserr` is set.

### DIFF
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
@@ -67,7 +67,7 @@ public class Utility {
    */
   public static void executeCommand(Config config, String command) {
     try {
-      ProcessBuilder pb = new ProcessBuilder(command);
+      ProcessBuilder pb = new ProcessBuilder("/bin/sh", "-c", command);
       if (config.redirectBuildOutputToStdErr) {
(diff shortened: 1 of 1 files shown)


msridhar:
Do we have any regression tests for this, or have you been able to test manually?

msridhar (approved):

PR #208, opened 2023-07-26
Update tests to compute distinct build command for target and downstream depenedencies

nimakarimipour:
This PR updates tests infrastructure to compute distinct commands for building target and downstream dependencies. This will reveal a bug in the implementation which is also fixed in this PR.


lazaroclapp (approved):
LGTM!

msridhar (approved):

PR #211, opened 2023-10-27
Index flow of nullable back to upstream through a field write

### LINKED ISSUE #209: Crash by Precondition - Missing triggered fixes from downstream
**Describe the bug**
Annotator incorrectly considers some fixes to be safe to inject on target. It currently considers fixes that introduce new errors on downstream dependencies that can be resolved via a separate fix on target, safe. It fails to also ensure that the resolving fix is also included in the fix tree.

**To Reproduce**
```java
@Test
  public void upperBoundCountForResolvableErrorOnDownstreamTest() {
    coreTestHelper
        .onTarget()
        .withSourceLines(
            "Bar.java",
            "package test;",
            "public class Bar {",
            "   public String foo;",
            "   public String foo2;",
            "   public void setFoo(String foo) {",
            "     this.foo = foo;",
            "   }",
            "   public String getFoo() {",
            "     return foo;",
            "   }",
            "}")
        .withDependency("Dep")
        .withSourceLines(
            "Dep.java",
            "package test.dep;",
            "import test.Bar;",
            "public class Dep {",
            "   public Bar bar = new Bar();",
            "   public void exec() {",
            "     bar.foo2 = bar.getFoo();",
            "   }",
            "}")
        .withExpectedReports()
        .disableBailOut()
        .enableDownstreamDependencyAnalysis(AnalysisMode.STRICT)
        .toDepth(5)
        .start();
  }
```
**Expected behavior**
Even though errors due to fix on `getFoo()` is resolvable on downstream, the corresponding fix is not included in the fix tree, hence, Annotator should consider this fix as unsafe.

**Stack trace**
```
java.lang.IllegalArgumentException
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:131)
	at edu.ucr.cs.riple.core.AnalysisMode$2.lambda$tag$0(AnalysisMode.java:69)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at edu.ucr.cs.riple.core.AnalysisMode$2.tag(AnalysisMode.java:61)
	at edu.ucr.cs.riple.core.Annotator.executeNextIteration(Annotator.java:150)
	at edu.ucr.cs.riple.core.Annotator.annotate(Annotator.java:110)
	at edu.ucr.cs.riple.core.Annotator.start(Annotator.java:72)
...
```

**Please complete the following information:**
 - Annotator Version: `1.3.9-SNAPSHOT`

**Additional context**
None

nimakarimipour:
This PR is built on (~#210~ landed)

Resolves #209 by indexing flow of nullable back to upstream through field writes. Previously we only indexed flow of nullable back to upstream via parameter passing, e.g. code below:
```java
class Target{
    Object getFoo()
    void bar(Object param);
}
class Dependency {
    Target t;
    void baz(){
       t.bar(t.getFoo());
    }
}
```
We know that making `getFoo()` will trigger a fix on `bar()` parameter. This PR updates Annotator to index all flows back to upstream including field writes. Hence, in the code below:
```java
class Target{
    Object bar;
    Object getFoo();
}
class Dependency {
    Target t;
    void baz(){
       t.bar = t.getFoo();
    }
}
```
Annotator now indexes the flow of nullable back to target through write to a field and includes the corresponding fix in the processing fix tree.

Please note this PR only updates indexing flow of nullable back to upstream. Annotator still does **not** index the impact of making public fields `@Nullable` on downstream dependencies.


msridhar:
> Please note this PR only updates indexing flow of nullable back to upstream. Annotator still does **not** index the impact of making public fields `@Nullable` on downstream dependencies.

I don't understand the importance of this.  Can not indexing the impact of making public fields `@Nullable` lead to similar bugs?

msridhar (approved):
I did not deeply understand the changes, but from what I can tell, this bug was fixed mostly by deleting code and a bit of generalization, which is great!

nimakarimipour:
> > Please note this PR only updates indexing flow of nullable back to upstream. Annotator still does **not** index the impact of making public fields `@Nullable` on downstream dependencies.
> 
> I don't understand the importance of this. Can not indexing the impact of making public fields `@Nullable` lead to similar bugs?

@msridhar The bug here was that we were flowing back to upstream through field writes and Annotator was not extending the tree accordingly. Making fields `@Nullable` on upstream can potentially cause problems in downstream dependencies but Annotator is not aware of that and will not enter an unexpected state. Please note in strict mode, we do not re-build all downstream dependencies to ensure no new bug has been introduced. We on rely on the errors indexed in the downstream scan phase which happens prior to annotating process.

PR #213, opened 2023-11-17
Updates method signature serialization for Nonull formal parameters

### LINKED ISSUE #212: Wrong serialization of @NotNull formal parameters with type use annotation
**Describe the bug**
Scanner serializes a wrong method signature for `@NotNull` formal parameters annotated with a type use annotation. The type use annotation fully qualified name is included in the serialized method signature which should not.

**To Reproduce**
Run scanner for the code below:
```java
package test;
import org.jetbrains.annotations.NotNull;
public class A {
  static void foo(@NotNull Object o) {}
  static void bar() { foo(null); }
}
```

Scanner serializes `foo(@org.jetbrains.annotations.NotNull java.lang.Object)` as the method signature for `o`.

**Expected behavior**
It should just serialize `foo(java.lang.Object)`.

**Stack trace**
Not applicable.

- Annotator version : `1.3.10-SNAPSHOT`

nimakarimipour:
This PR fixes #212 by updating method signature serialization for nonnull formal parameters. It reuses the existing serialization methods for symbols that ensure the correct method signature is printed.

### DIFF
--- a/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
+++ b/annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
@@ -558,7 +558,7 @@ public class CoreTest extends AnnotatorBaseCoreTest {
   }
 
   @Test
-  public void nonnullTypeUseAcknowledgmentTest() {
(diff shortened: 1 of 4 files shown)


msridhar (approved):
LGTM!

PR #214, opened 2023-11-17
Updates symbol serialization in all Location subclasses in Scanner

nimakarimipour:
Followup to #213, this PR fixes symbol serialization in all `Location` subclasses.

### DIFF
--- a/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/location/MethodLocation.java
+++ b/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/location/MethodLocation.java
@@ -23,6 +23,8 @@
 package edu.ucr.cs.riple.scanner.location;
 
 import com.sun.tools.javac.code.Symbol;
+import edu.ucr.cs.riple.scanner.Serializer;
+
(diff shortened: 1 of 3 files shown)


annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/location/MethodLocation.java
```diff
@@ -45,7 +46,7 @@ public String tabSeparatedToString() {
         "\t",
         type.toString(),
         enclosingClass.flatName(),
```
msridhar, line 48: Should we do this, for consistency?
```suggestion
        Serializer.serializeSymbol(enclosingClass),
```

nimakarimipour: Sure, we can. I will update it for all locations.

nimakarimipour: @msridhar Updated in all subclasses, looks much better now.

msridhar (approved):
👍

PR #215, opened 2023-11-30
chore(readme): moves optional agrs to own file for readability

raghuganapathyUCR:
@nimakarimipour  I moved the optional args to OPTIONS.md, only keeping the ones that are most useful to the user on the Readme. Let me know if more changes are needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,8 @@ This sections describes how to run `Annotator` on any project.
 - ### Requirements for the Target Project
 
   #### Dependencies
-	-  `NullAway` checker must be activated with version >= `0.10.10`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -239,7 +226,7 @@ In this approach, you will initialize all flag values in a single file and pass
   "INFERENCE_ACTIVATION": true
 }
 ```
-Here is a template command you can use to run Annotator from the CLI, using config.json - 
+Here is a template command you can use to run Annotator from the CLI, using config.json -
```
nimakarimipour, line 229: This part needs to be removed as well.

nimakarimipour (changes requested):

PR #220, opened 2023-12-17
Update Library Model Loader to mark fields Nullable

nimakarimipour:
Followup on an update in NullAway recent change [#878](https://github.com/uber/NullAway/pull/878) which enables NullAway to mark fields as `@Nullable`. This PR updates our `LibraryModelsLoader` to enable communication of Annotator and NullAway to mark fields `@Nullable` while analyzing downstream dependencies. This is required to prepare the code for the upcoming change which enables Annotator to be informed of impacts of making fields `@Nullable` on downstream dependencies.


annotator-core/src/main/java/edu/ucr/cs/riple/core/injectors/VirtualInjector.java
```diff
@@ -75,33 +82,57 @@ public void injectAnnotations(Set<AddAnnotation> changes) {
...
+            // we have to consider both a and b be nullable and write each one
+            // ("Foo\ta" and "Foo\tb")
+            // on a separate line.
+            annot.getLocation().toField().variables.stream()
+                .map(variable -> annot.getLocation().clazz + "\t" + variable + "\n"));
+  }
+
+  /**
+   * Writes the passed annotation to the passed file. It uses the passed mapper to map the
+   * annotation to a string. And writes each string to a separate line.
```
msridhar, line 112:
```suggestion
   * annotation to a stream of strings. And writes each string to a separate line.
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/injectors/VirtualInjector.java
```diff
@@ -75,33 +82,57 @@ public void injectAnnotations(Set<AddAnnotation> changes) {
...
+   *
+   * @param annotations Annotations to be written.
+   * @param path Path to the file to be written.
+   * @param mapper Mapper to map the annotation to a string.
```
msridhar, line 116:
```suggestion
   * @param mapper Mapper to map the annotation to a stream of strings.
```

annotator-core/src/main/java/edu/ucr/cs/riple/core/injectors/VirtualInjector.java
```diff
@@ -75,33 +82,57 @@ public void injectAnnotations(Set<AddAnnotation> changes) {
...
+  private static void writeAnnotationsToFile(
+      Stream<AddAnnotation> annotations,
+      Path path,
+      Function<AddAnnotation, Stream<String>> mapper) {
+    try (BufferedOutputStream os = new BufferedOutputStream(new FileOutputStream(path.toFile()))) {
+      Set<String> rows = annotations.flatMap(mapper).collect(Collectors.toSet());
       for (String row : rows) {
         os.write(row.getBytes(Charset.defaultCharset()), 0, row.length());
       }
       os.flush();
```
msridhar, line 127: Not a big deal, but I don't think this call is needed, since the stream will immediately be closed afterward

annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
```diff
@@ -82,6 +83,23 @@ public static void executeCommand(Config config, String command) {
     }
   }
 
+  /**
+   * Clears the content of the file at the given path if the file exists.
```
msridhar, line 87: Why don't you just delete the file?

library-model-loader/src/main/java/edu/ucr/cs/riple/librarymodel/LibraryModelLoader.java
```diff
@@ -38,21 +41,28 @@
...
-   * Loads a file from resources and parses the content into set of {@link
-   * com.uber.nullaway.LibraryModels.MethodRef}.
+   * Loads a file from resources and creates an instance of type T from each line of the file.
    *
    * @param name File name in resources.
-   * @return ImmutableSet of content in the passed file. Returns empty if the file does not exist.
+   * @return ImmutableSet of contents in the file. Returns empty if the file does not exist.
    */
-  private ImmutableSet<MethodRef> parseTSVFileFromResourcesToMethodRef(String name) {
+  private <T> ImmutableSet<T> parseTSVFileFromResourcesToMemberRef(
```
msridhar, line 64: Not for this PR, but, if in NullAway we had a common marker interface `MemberRef` for `MethodRef` and `FieldRef`, then `MemberRef` could be an upper bound for `T`, giving slightly more type safety

library-model-loader/src/main/java/edu/ucr/cs/riple/librarymodel/LibraryModelLoader.java
```diff
@@ -120,4 +130,29 @@ public ImmutableSet<MethodRef> nonNullReturns() {
...
+
+  @Override
+  public ImmutableSet<FieldRef> nullableFields() {
+    return nullableFields;
+  }
+
+  @Override
+  public ImmutableList<StreamTypeRecord> customStreamNullabilitySpecs() {
+    return LibraryModels.super.customStreamNullabilitySpecs();
+  }
```
msridhar, line 142: Isn't this override unnecessary?  It just calls the super method?

library-model-loader/src/main/java/edu/ucr/cs/riple/librarymodel/LibraryModelLoader.java
```diff
@@ -120,4 +130,29 @@ public ImmutableSet<MethodRef> nonNullReturns() {
...
+
+  /**
+   * Factory interface for creating an instance of type T from a string array.
+   *
+   * @param <T> Type of the instance to create.
+   */
+  interface Factory<T> {
```
msridhar, line 149: As noted above, we could eventually have greater type safety here if we could upper bound `T` (not for this PR)

msridhar (commented):
Overall looks good!  I have a few comments.  Also, why is `nullable-methods.tsv` checked in as an empty file?

annotator-core/src/main/java/edu/ucr/cs/riple/core/injectors/VirtualInjector.java
nimakarimipour: Sure, fixed in  f775354be1a90e45c7f67ffa5cdb0ecfe7010488

annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
nimakarimipour: Sure, that sounds better. Fixed in 6dd203efc9ca353952a2e8768243772a626ddf8c

library-model-loader/src/main/java/edu/ucr/cs/riple/librarymodel/LibraryModelLoader.java
nimakarimipour: Thanks for catching this, I really don't know what I was thinking 🤦🏻‍♂️ de61967d01ff7e45865ce4f03a9cd42c63e3a2f2

library-model-loader/src/main/java/edu/ucr/cs/riple/librarymodel/LibraryModelLoader.java
nimakarimipour: Sounds good. Added the issue #222.

library-model-loader/src/main/java/edu/ucr/cs/riple/librarymodel/LibraryModelLoader.java
nimakarimipour: Thanks, created the corresponding issue #222.

nimakarimipour:
> Overall looks good! I have a few comments. Also, why is `nullable-methods.tsv` checked in as an empty file?

@msridhar Thank you very much for the review, this is ready for another round.

Regarding
> Also, why is nullable-methods.tsv checked in as an empty file?

We weren't creating / deleting files in resource directory (which we are now 6dd203e). We used to assume the file exists and clear contents instead of deleting. With the new approach, we no longer need the file.

annotator-core/src/main/java/edu/ucr/cs/riple/core/injectors/VirtualInjector.java
msridhar: I still see the call to `os.flush()`?  Why is this needed?

nimakarimipour: Is the call to `os.flush()` redundant (or guaranteed to be executed after the block) in `try-with-resources` block ? IntelliJ only grays out call to `os.close()` but not for `os.flush()`.

msridhar: `try-with-resources` guarantees the stream will be closed, and I believe it will be flushed automatically before it is closed.

nimakarimipour: Ok thanks for that. Then we can remove it. Fixed in 24f0a244677677c0cc7a15a154eda18bd8425080

msridhar (approved):
The PR still includes an empty `nullable-methods.tsv` file.  After removing that file from the PR it looks good to go

nimakarimipour:
@msridhar I don't see the `nullable-methods.tsv` file on my branch.  Isn't this PR actually deleting the existing `nullable-methods.tsv` on master branch ? Please let me know if I am missing something.

msridhar:
> @msridhar I don't see the `nullable-methods.tsv` file on my branch. Isn't this PR actually deleting the existing `nullable-methods.tsv` on master branch ? Please let me know if I am missing something.

Yes, you're right.  For some reason GitHub shows the change strangely.  Locally git shows the file as removed.  Sorry for the confusion.

nimakarimipour:
@msridhar Ok great, will land this one now and update #221.

PR #221, opened 2023-12-20
Add support for indexing of impacts of changes on fields on downstream dependencies

nimakarimipour:
This PR adds support for indexing impact of changes on fields on downstream dependencies. Prior to this PR, Annotator might make a field `@Nullable` that can trigger errors on downstream dependencies, with this PR, Annotator is aware of these impacts and if changes on fields, create an unresolvable error on downstream dependencies, the containing fix tree is **rejected**.

See example below:

#### On Target
```java
public class Foo {

  public Object f0;

  public Object f1;

  public Object f2 = new Object();
}
```
#### On Dependency (Dep)
```java
package test.depa;

import test.target.Foo;

public class Dep {

    public void directAccess(Foo foo){
         // making f0 will create an unresolvable error here.
         foo.f0.toString();
    }

    public void flowF1BackToF2(Foo foo) {
      // If f1 is annotated as @Nullable, the corresponding error on f2 can be resolved on target.
      foo.f2 = foo.f1;
    }
}
```

Prior to this PR, Annotator would have made `f0` `@Nullable` even thought it causes an unresolvable error in `Dep` module. With this PR, 
 - Annotator is aware of this impact and will avoid making `f0` `@Nullable`
 - Annotator is aware that the triggered error by making `f1` `@Nullable` is resolvable by making `f2` `@Nullable`, therefore it includes the corresponding fix (`f2`) in any tree containing `f1`.


annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpactCacheImpl.java
```diff
@@ -80,22 +82,34 @@ public DownstreamImpactCacheImpl(Context context) {
    */
   private ImmutableSet<Location> retrieveLocationsToCacheImpactsOnDownstreamDependencies(
       Context context, ModuleInfo moduleInfo) {
-    // Collect callers of public APIs in module.
+    ImmutableSet.Builder<Location> locationsToCache = ImmutableSet.builder();
+    // Collect public methods with non-primitive return types.
+    // Used to collect callers of each method.
```
msridhar, line 87: Should this comment be before line 90?

annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/ClassFieldRecord.java
```diff
@@ -96,6 +96,16 @@ public void addFieldDeclaration(FieldDeclaration fieldDeclaration) {
     this.fields.add(new FieldDeclarationRecord(fieldDeclaration));
   }
 
+  /**
+   * Checks if the class contains a field with the given name.
```
msridhar, line 100: This method is doing something different than the docs.  It seems to look for a single field declaration that declares _all_ the names in the `Set` given as a parameter.  We should update the docs and consider renaming the method.

annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
```diff
@@ -89,14 +88,12 @@ public static void executeCommand(Config config, String command) {
...
-          Files.newBufferedWriter(
-              path, Charset.defaultCharset(), StandardOpenOption.TRUNCATE_EXISTING)) {
-        writer.flush();
-      } catch (IOException e) {
-        throw new RuntimeException("Could not clear content of file: " + path, e);
-      }
+    try {
+      Files.deleteIfExists(path);
+      Files.createDirectories(path.getParent());
+      Files.createFile(path);
```
msridhar, line 94: I think this is modifying logic introduced in #220.  Still not sure why we are creating an empty file here.

msridhar (commented):
Overall looks really good!  Just a couple of comment

annotator-core/src/main/java/edu/ucr/cs/riple/core/cache/downstream/DownstreamImpactCacheImpl.java
nimakarimipour: Thanks for catching this. Fixed it in 15c8e1ec3b98b0e8b4ce4d26f2c2b83705e72206

annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java
nimakarimipour: Yes, this is fixed in #220 now.

annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/ClassFieldRecord.java
nimakarimipour: @msridhar Yes, I agree. Regarding field declarations (or `OnFIeld` instances) we have to document somewhere how they are created. Throughout the entire `core` module, every `OnField` instance contains all inline field names for the declaration, for instance for a declaration `Foo a, b, c`, all `OnField` instances targeting this declaration, contain all `a`, `b` and `c` on their `variables` property. This has been done at deserializing fixes [here](https://github.com/ucr-riple/NullAwayAnnotator/blob/f773208495a0e7a8ee2432a8f3b0da4ddb0da7a1/annotator-core/src/main/java/edu/ucr/cs/riple/core/checkers/nullaway/NullAway.java#L128). That is the reason why I used `equals` instead of checking if it contains it. Will change the name and the documentation now.

nimakarimipour: Updated javadoc for this method 71b901df47ddf678fb4cfb7097e47a88be63b456.
Will work on a more detailed documentation with revisiting all methods working on fields in a followup PR.

annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/ClassFieldRecord.java
```diff
@@ -96,6 +96,25 @@ public void addFieldDeclaration(FieldDeclaration fieldDeclaration) {
...
+   *     Object a, b, c;
+   * }
+   * }</pre>
+   *
+   * The call with ["a", "b", "c"] will return true, but for ["a", "b"] or ["a"] will return false.
+   *
+   * @param name The set of names to be checked for a matching field declaration.
+   * @return true if there exists a field declaration with exactly the given names; false otherwise.
+   */
+  public boolean hasExactFieldDeclarationWithNames(Set<String> name) {
```
msridhar, line 114: Let's rename the parameter:
```suggestion
  public boolean hasExactFieldDeclarationWithNames(Set<String> names) {
```

msridhar (approved):
LGTM!  Just one very minor comment, can land after that

PR #227, opened 2024-02-14
Update parse method in Injector to handle null path as argument

raghuganapathyUCR:
@nimakarimipour To mitigate the Injector crashes observed in testing on micronaut, I have updated the parse method to explicitly return null when a NoSuchFileException occurs. 

I've tested these changes on micronaut and litiengine, observing successful annotator runs


injector/src/main/java/edu/ucr/cs/riple/injector/Injector.java
```diff
@@ -147,14 +148,20 @@ public Set<FileOffsetStore> removeAnnotations(Set<RemoveAnnotation> requests) {
    */
   @Nullable
   public static CompilationUnit parse(Path path) {
-    // Set parser configuration to Java 17.
+    // TODO: Figure out how to differentiate a generated file from a non-generated file.
```
nimakarimipour, line 151: We don't really need to differentiate a generated file from a non-generated file and this TODO might not be applicable here, would you please remove it. In both cases we would like to just return `null` and the caller should handle that.

injector/src/main/java/edu/ucr/cs/riple/injector/Injector.java
```diff
@@ -147,14 +148,20 @@ public Set<FileOffsetStore> removeAnnotations(Set<RemoveAnnotation> requests) {
    */
   @Nullable
   public static CompilationUnit parse(Path path) {
```
nimakarimipour, line 150: Can we make `path` nullable please ? `@Nullable Path path`

injector/src/main/java/edu/ucr/cs/riple/injector/Injector.java
```diff
@@ -147,14 +148,20 @@ public Set<FileOffsetStore> removeAnnotations(Set<RemoveAnnotation> requests) {
...
+    }
     ParserConfiguration parserConfiguration = new ParserConfiguration();
+    // Set parser configuration to Java 17.
     parserConfiguration.setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_17);
     StaticJavaParser.setConfiguration(parserConfiguration);
     try {
       return StaticJavaParser.parse(path);
     } catch (FileNotFoundException e) {
       return null;
+    } catch (NoSuchFileException e) {
```
nimakarimipour, line 163: Let's not use both `NoSuchFileException` and `FileNotFoundException` since they can be collapsed into one, and use only `NoSuchFileException` from `java.nio.file` package and remove the other one.

nimakarimipour (changes requested):
Thanks for the PR. Few comments.

nimakarimipour (approved):

PR #230, opened 2024-05-13
Optimize field registry population mechanism by parsing each source file only once

nimakarimipour:
In [FieldRegistryPopulation](https://github.com/ucr-riple/NullAwayAnnotator/blob/6c17b8bccb08517f5ce3249d3285405ceeb47748/annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/FieldRegistry.java#L97) we process a single compilation unit tree multiple times as this method receives a pair of a class flat name and a path to source file containing that class, hence, for compilation unit trees that contains multiple classes, inner classes and anonymous classes we parse the same source file multiple times. This PR updates this mechanism by caching the latest parsed source file. This optimization is according to the assumption that Scanner visits all classes within a single compilation unit tree consecutively.

### DIFF
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/FieldRegistry.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/FieldRegistry.java
@@ -94,18 +94,22 @@ public class FieldRegistry extends Registry<ClassFieldRecord> {
 
   @Override
   protected Builder<ClassFieldRecord> getBuilder() {
-    AtomicReference<Pair<Path, CompilationUnit>> lastParsed = new AtomicReference<>();
+    // This method is called with a pair of a class flat name and a path to source file containing
+    // the class. To avoid parsing a source file multiple times, we keep the last parsed
+    // source file in a reference. This optimization is according to the assumption that Scanner
+    // visits all classes within a single compilation unit tree consecutively.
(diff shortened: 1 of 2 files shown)


annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/FieldRegistry.java
```diff
@@ -92,46 +93,63 @@ protected void setup() {
...
+        // Path to class.
+        Path path = Helper.deserializePath(values[1]);
+        CompilationUnit tree;
+        if (lastParsedSourceFile.a != null && lastParsedSourceFile.a.equals(path)) {
+          // Already visited.
+          tree = lastParsedSourceFile.b;
+        } else {
+          tree = Injector.parse(path);
+        }
+        lastParsedSourceFile = new Pair<>(path, tree);
```
msridhar, line 115: Could this line be inside the preceding `else` block?

msridhar (approved):
LGTM, just one minor question, does not need to be addressed before landing

annotator-core/src/main/java/edu/ucr/cs/riple/core/registries/field/FieldRegistry.java
nimakarimipour: cfcfb35d70698ee0514e01e9faf0683df9a84872

PR #231, opened 2024-05-13
Skip serialization for generated elements in Scanner

nimakarimipour:
[WIP]

### DIFF
--- a/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
+++ b/annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
@@ -45,6 +45,8 @@ import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.util.Context;
+import edu.ucr.cs.riple.scanner.generatedcode.CodeAnnotationInfo;
 import edu.ucr.cs.riple.scanner.out.ClassRecord;
 import edu.ucr.cs.riple.scanner.out.ImpactedRegion;
 import edu.ucr.cs.riple.scanner.out.MethodRecord;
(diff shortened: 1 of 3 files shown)


annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/AnnotatorScanner.java
```diff
@@ -122,14 +131,15 @@ public Description matchNewClass(NewClassTree tree, VisitorState state) {
     if (methodSymbol == null) {
       throw new RuntimeException("not expecting unresolved method here");
     }
-    if (methodSymbol.owner.enclClass().getSimpleName().isEmpty()) {
+    if (methodSymbol.owner.enclClass().getSimpleName().isEmpty()
+        || isInGeneratedClass(methodSymbol, state.context)) {
       // An anonymous class cannot declare its own constructors, so we do not need to serialize it.
```
msridhar, line 136: comment needs an update?

annotator-scanner/src/main/java/edu/ucr/cs/riple/scanner/generatedcode/CodeAnnotationInfo.java
```diff
@@ -0,0 +1,119 @@
...
+            : ASTHelpers.enclosingClass(symbol);
+    if (classSymbol == null) {
+      return false;
+    }
+    Symbol.ClassSymbol outermostClassSymbol = get(classSymbol);
+    return hasDirectAnnotationWithSimpleName(outermostClassSymbol, "Generated");
+  }
+
+  /**
+   * Retrieve the (outermostClass, isNullMarked) record for a given class symbol.
```
msridhar, line 88: I think this copy-pasted comment is wrong, since the code around null-markedness has been removed

msridhar (commented):
Couple of minor comments.  But I have a higher-level question.  Technically, whether NullAway checks `@Generated` code or not is configurable via flags, see [here](https://github.com/uber/NullAway/wiki/Configuration#treat-generated-as-unannotated).  In fact, by default, I think NullAway _will_ check code inside `@Generated` classes.  This PR seems to unconditionally assume that `@Generated` code will not be checked, is this correct?  If so, we might want to make this configurable via a flag, and probably that flag should have the same default as NullAway itself.

nimakarimipour:
@msridhar Yes this PR unconditionally skips all serializations in generated classes. I will work on adding a controlling option for that. I think this should be an option that is provided to Annotator, and Annotator pass that to Scanner in config file rather than directly providing to Scanner.  But what I am not sure of is that how we can describe this option to avoid confusion. Anything similar to `ignore-generated` is a bit confusing since it is not really up to Annotator, if this option is provided to Annotator and NullAway still suggest a fix on a generated class, we will still modify that.  Another alternative is that we update NullAway serialization that it serialize how it treats generated code in the same file it serializes its serialization version, then Annotator read that file. This way Annotator is just following the configuration on NullAway. Would you please let me know what you think of this ?

msridhar:
@nimakarimipour I think it makes sense for NullAway for serialize the options it used to handle generated code; this would certainly be easiest.  But to pass this information to the scanner, we would need to run NullAway once _before_ running the scanner, correct?  Not sure what order things are run right now.

Taking a step back, #230 seems to have yielded a big performance improvement for the use case we cared about.  So maybe this change can be lower priority?

PR #232, opened 2024-05-14
Add a language level option for parsing

msridhar:
Needed to work around https://github.com/javaparser/javaparser/issues/4041.  Users impacted by that issue can set the language level to 11, assuming they are not using language constructs introduced after that version.


msridhar:
One thought here.  Since we have already released a version of the annotator with the default language level as 17, switching the default to 11 would be a breaking change.  Should we stick with the default level of 17 maybe?  I think that would work fine in nearly all case; as of now, we only need configurability here because of https://github.com/javaparser/javaparser/issues/4041.  At Uber we could easily just configure the level to be 11.

PR #233, opened 2024-06-10
Optimize memory usage for FieldDeclarationNode

nimakarimipour:
This PR improves the memory usage of `FieldDeclarationNode` instances, significantly reducing the memory allocation during the execution of Annotator. The main change involves replacing the reference to `DeclarationNode` with boolean variables, which store the necessary information previously retrieved through the `DeclarationNode` reference.


msridhar (approved):
Nice fix!

PR #236, opened 2024-07-15
Update to Gradle 8.9

msridhar:
Just to stay up to date.  Note that builds do not work if the JVM used to run Gradle is more recent than JDK 11.  I think we need to switch our formatting check to spotless to get more recent JVMs to work.  But that can be separate.


nimakarimipour:
> Just to stay up to date. Note that builds do not work if the JVM used to run Gradle is more recent than JDK 11. I think we need to switch our formatting check to spotless to get more recent JVMs to work. But that can be separate.

Thank you very much for this, landing it now. Sure will change to `spotless` in a separate PR.

Issue #238, opened 2024-07-21
Problems with Directory Handling for -d,--dir Flag in NullAway Annotator


akbarkanso:
**Describe the bug**

I encountered issues when using the required -d,--dir <arg> flag to specify the directory for output serialization in NullAway Annotator. The tool expects a nonnull_elements.tsv file in the directory, but handling this file results in various exceptions.

**To Reproduce**

- Run the NullAway Annotator with the Directory:
`java -jar /path/to/annotator-core-1.3.14.jar -d "nullawayResults" .....`

- Encounter NoSuchFileException:
`java.nio.file.NoSuchFileException: nullawayResults/0/nonnull_elements.tsv`

- Create the missing File:
`touch nullawayResults/0/nonnull_elements.tsv`

- Run the NullAway Annotator Again:
`java -jar /path/to/annotator-core-1.3.14.jar -d "nullawayResults" .....`

- Encounter DirectoryNotEmptyException:
`java.nio.file.DirectoryNotEmptyException: nullawayResults/0`

**Expected behavior**

The tool should either:
1. Create the nonnull_elements.tsv file if it does not exist, or
2. Properly handle an empty directory without requiring pre-existing files

**Stack trace**

Stack trace for DirectoryNotEmptyException:
```
Exception in thread "main" java.lang.RuntimeException: java.nio.file.DirectoryNotEmptyException: nullawayResults/0
	at edu.ucr.cs.riple.core.module.ModuleConfiguration.<init>(ModuleConfiguration.java:74)
	at edu.ucr.cs.riple.core.Config.lambda$new$0(Config.java:418)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at edu.ucr.cs.riple.core.Config.<init>(Config.java:420)
[...]

Issue #237, opened 2024-07-15
Detect compiler failures and print an appropriate error message in such cases


msridhar:
I believe whenever the annotator runs to either build the target itself or to build downstream deps, it expects the compilation run to _succeed_, i.e., complete without compilation errors.  (NullAway needs to be run at the WARNING level for this to occur.)  If this is true, then annotator should detect when a build command fails, print an appropriate error message, and terminate the run.  Right now, the annotator ignores build failures, which often leads to an incrutable failure later in the annotator run, like crashing due to the absence of some expected tsv file.

This could be implemented by directly checking the return code of the executed command in [`executeCommand`](https://github.com/ucr-riple/NullAwayAnnotator/blob/cb099624831daa782a1f03c52b1c5d01d3dd829a/annotator-core/src/main/java/edu/ucr/cs/riple/core/util/Utility.java#L68-L68) or at callers of this function if that's better.

PR #239, opened 2024-08-08
Add check for scanner output

nimakarimipour:
This PR introduces a check on the `AnnotatorScanner` output and notifies the user if `AnnotatorScanner` is incorrectly configured or has crashed. Before this PR, the Annotator would simply crash with a 'nonnull_elements.tsv not found' error (generated by AnnotatorScanner), which did not clearly indicate the underlying issue.

This PR is also related to #238.


msridhar (approved):
LGTM. I think addressing #237 at some point would also be helpful

PR #258, opened 2024-11-07
Add test cases for annotation insertion/deletion on array component type.

avenger2597:
This PR adds test cases for the addition and deletion of annotations on the component type of arrays. It includes tests to verify successful annotation updates for arrays containing both primitive types and wrapper class types.


nimakarimipour (approved):
LGTM!

PR #267, opened 2025-02-12
Use Type Use Annotations correctly when assigning null to non-null array component.

avenger2597:
This PR adds Nulllable annotation at the component of an array in case when null is being assigned to it. This feature uses TypeUseAnnotations and is available only in Jspecify mode. We check for a specific error message from Nullaway checker - `ASSIGN_NULLABLE_TO_NONNULL_ARRAY` and based on this create Type Use annotation at the required index.


annotator-core/src/main/java/edu/ucr/cs/riple/core/checkers/nullaway/NullAway.java
```diff
@@ -134,7 +135,14 @@ private NullAwayError deserializeErrorFromTSVLine(ModuleInfo moduleInfo, String
     if (nonnullTarget == null) {
       annotations = Set.of();
     } else if (Utility.isTypeUseAnnotation(config.nullableAnnot)) {
-      annotations = Set.of(new AddTypeUseMarkerAnnotation(nonnullTarget, config.nullableAnnot));
+      if (errorType.equals(NullAwayError.ASSIGN_NULLABLE_TO_NONNULL_ARRAY)) {
+        annotations =
+            Set.of(
+                new AddTypeUseMarkerAnnotation(
```
nimakarimipour, line 141: add a comment here that [1, 0] is targeting array's component type.

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -572,4 +573,32 @@ public void nonnullTypeUseAnnotationOnFormalParameterAcknowledgmentTest() {
     // No annotation should be added even though it can reduce the number of errors.
     Assert.assertEquals(coreTestHelper.getLog().getInjectedAnnotations().size(), 0);
   }
+
+  @Test
+  public void fieldAssignNullableNonNullArrayContent() {
```
nimakarimipour, line 578: The test case name is a bit unclear, maybe `assignNullableToNonnullArrayComponentTypeTest` ?

nimakarimipour (changes requested):
Some notes.

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
```diff
@@ -572,4 +573,33 @@ public void nonnullTypeUseAnnotationOnFormalParameterAcknowledgmentTest() {
...
+            "     arr[0] = null;",
+            "   }",
+            "}")
+        .withExpectedReports(
+            new TReport(
+                new OnField("Main.java", "test.Main", Set.of("arr")),
+                ImmutableList.of(ImmutableList.of(1, 0)),
+                -1))
+        .setPredicate(
+            (expected, found) ->
```
nimakarimipour, line 596: I don't adding a custom predicate here is necessary. Is there a reason you added this ?

nimakarimipour:
@avenger2597 I added the logic for enabling jspecify mode only by request for unit tests.

annotator-core/src/main/java/edu/ucr/cs/riple/core/checkers/nullaway/NullAway.java
avenger2597: Added a comment explaining this

annotator-core/src/test/java/edu/ucr/cs/riple/core/CoreTest.java
avenger2597: Renamed the test case and output file.

annotator-core/src/main/java/edu/ucr/cs/riple/core/checkers/nullaway/NullAway.java
```diff
@@ -134,7 +135,16 @@ private NullAwayError deserializeErrorFromTSVLine(ModuleInfo moduleInfo, String
     if (nonnullTarget == null) {
       annotations = Set.of();
     } else if (Utility.isTypeUseAnnotation(config.nullableAnnot)) {
-      annotations = Set.of(new AddTypeUseMarkerAnnotation(nonnullTarget, config.nullableAnnot));
+      if (errorType.equals(NullAwayError.ASSIGN_NULLABLE_TO_NONNULL_ARRAY)) {
+        // The typeIndex ImmutableList.of(1, 0) specifies that the annotation
```
nimakarimipour, line 139: Perhaps just `[1, 0]` instead of `ImmutableList.of(1, 0)`, the latter feels unnecessarily verbose for conveying simple information. The purpose of this comment is merely to indicate that `[1, 0]` targets the component type of an array, while the specific type itself is irrelevant.

I would phrase it as:
```
The error `ASSIGN_NULLABLE_TO_NONNULL_ARRAY` from NullAway triggers a fix on an array variable, with `[1, 0]` indicating its component type.
```

nimakarimipour (changes requested):
Looking great, just one change and ready to land.

annotator-core/src/main/java/edu/ucr/cs/riple/core/checkers/nullaway/NullAway.java
avenger2597: Updated the comment to this

nimakarimipour (approved):

PR #274, opened 2025-09-21
 Upgrade javaparser to version 3.27.0

akbarkanso:
Upgrade javaparser to version 3.27.0. Fixes the issue with Java 21 case statement using a record pattern or a type pattern


msridhar:
@nimakarimipour could you take a look?

PR #276, opened 2026-03-29
Add Java 21 language level support to -ll flag and document it

akbarkanso:
## Problem

The `-ll` / `--language-level` flag accepts only `"11"` and `"17"`.

Projects using Java 21 features (record patterns, switch pattern matching—stable since JEP 440/441) cause a `ParseException` before any annotation inference begins:

```

ParseException: Record patterns are not supported. Pay attention that this
feature is supported starting from 'JAVA_21' language level.

```

The bundled JavaParser already has `ParserConfiguration.LanguageLevel.JAVA_21`, but `Config.getLanguageLevel()` throws an `IllegalArgumentException` before it is ever reached.

Additionally, `-ll` is fully absent from both `README.md` and `OPTIONS.md`, making it invisible to users.

## Changes

- `Config.java`: add `case "21": return JAVA_21` and update the CLI help string.
- `OPTIONS.md`: document `-ll` in the optional flags table with default and supported values.
- `README.md`: add `-ll` to the short "useful optional arguments" table near the quick-start command example.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -98,4 +98,18 @@ jobs:
           ANNOTATOR_TEST_DISABLE_PARALLEL_PROCESSING: "false"
           ANNOTATOR_TEST_DISABLE_CACHING: "false"
         run: ./gradlew :annotator-core:test --tests "edu.ucr.cs.riple.core.Java17Test"
+  parser-configuration-21:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 21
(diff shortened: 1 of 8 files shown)


msridhar:
@akbarkanso thank you for this contribution!  Can you add at least one test for Java 21 features so we have something to check this is working?

akbarkanso:
@msridhar  I've added a test for Java 21 features (record patterns) along with a dedicated CI job to run it under JDK 21

msridhar (approved):
LGTM, thanks!  @nimakarimipour I'm going to merge but if you have any concerns we can fix in a follow up

Issue #240, opened 2024-08-14
Parser fails when it encounters a variable named `_`


donalmurtagh:
**Describe the bug**
In JDK v22, a feature was added to the Java language that allows [unused variables](https://docs.oracle.com/en/java/javase/22/language/unnamed-variables-and-patterns.html) to be named `_`. When the java parser encounters such a variable, it fails

**To Reproduce**
Run the annotator under JDK v22 on a codebase that contains an unused variable named `_`. The following sections show the relevant dependency versions, configuration, etc.

### build.gradle

Relevant sections of the build file are

```groovy

plugins {
    id "net.ltgt.errorprone" version "4.0.1"
}

dependencies {
    implementation "org.jspecify:jspecify:1.0.0"
    annotationProcessor 'edu.ucr.cs.riple.annotator:annotator-scanner:1.3.14'
    annotationProcessor "com.uber.nullaway:nullaway:0.11.1"
    errorprone "com.google.errorprone:error_prone_core:2.30.0"
    errorproneJavac "com.google.errorprone:javac:9+181-r4173-1"
}

tasks.withType(JavaCompile).configureEach {
    options.errorprone {
        check("NullAway", CheckSeverity.ERROR)
        check("AnnotatorScanner", CheckSeverity.ERROR)
        option("NullAway:AnnotatedPackages", "com.myapp")
        option("NullAway:SerializeFixMetadata", "true")
        option("NullAway:FixSerializationConfigPath", "/Users/donal/tmp/annotator-output/nullaway.xml")
        option("AnnotatorScanner:ConfigPath", "/Users/donal/tmp/annotator-output/scanner.xml")
    }

    options.compilerArgs << "-Xmaxerrs"<< "100000"
    options.compilerArgs << "-Xmaxwarns" << "100000"
}
```

### paths.tsv
A tab character separates the two paths.
```
/Users/donal/tmp/annotator-output/nullaway.xml	/Users/donal/tmp/annotator-output/scanner.xml
```

### Annotator Command
```
java -jar ~/tmp/annotator-core-1.3.14.jar -d ~/tmp/annotator-output -i jakarta.annotation.PostConstruct -bc "cd ~/workspace/my-app && ./gradlew build -x test" -cn NULLAWAY -cp ~/tmp/paths.tsv -n org.jspecify.annotations.Nullable
```

**Expected behavior**
Annotator should run successfully

**Stack trace**
```
[...]

PR #279, opened 2026-08-07
Accept every language level the bundled parser supports

AdzerKI:
## Problem

`-ll` accepts only `11`, `17` and `21`. A project compiled with a newer release fails to parse before any inference runs — on JDK 22+ an unnamed variable aborts the whole run:

```
ParseException: javaparser was not able to parse file at: .../ClickHouseBatchInserter.java
Parse problem:(line 51,col 28) '_' is a reserved keyword.
```

That is #240. Unnamed variables are stable since JEP 456 (JDK 22), so the failure is reachable from any current LTS, and there is no `-ll` value that avoids it: the bundled JavaParser 3.27.0 tops out at `JAVA_21`.

Each new release otherwise needs another `case` in `Config`, as #276 did for 21.

## Changes

- `Config.getLanguageLevel` resolves the value against `ParserConfiguration.LanguageLevel` instead of a fixed switch, so every level the parser knows is accepted and future releases need no code change. Unknown values still fail with the same `IllegalArgumentException`.
- JavaParser 3.27.0 → 3.28.2, which adds `JAVA_22` … `JAVA_26`.
- `-ll` docs in `README.md` and `OPTIONS.md` state the range and why the value must match the target project.

## Verification

Built from this branch and run with `-ll 25` over a 21-module Gradle project on JDK 25 (records, switch patterns, unnamed variables): the parse failure is gone and inference completes.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -112,4 +112,18 @@ jobs:
           ANNOTATOR_TEST_DISABLE_PARALLEL_PROCESSING: "false"
           ANNOTATOR_TEST_DISABLE_CACHING: "false"
         run: ./gradlew :annotator-core:test --tests "edu.ucr.cs.riple.core.Java21Test"
+  parser-configuration-25:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 25
+        uses: actions/setup-java@v4
+        with:
+          java-version: 25
+          distribution: 'temurin'
+      - name: Build with Gradle
+        env:
+          ANNOTATOR_TEST_DISABLE_PARALLEL_PROCESSING: "false"
+          ANNOTATOR_TEST_DISABLE_CACHING: "false"
+        run: ./gradlew :annotator-core:test --tests "edu.ucr.cs.riple.core.Java25Test"
 
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -20,4 +20,4 @@ Here are all __optional__ arguments that can alter Annotator's default configura
 | `-drdl, --deactivate-region-detection-lombok`          | Deactivates region detection for Lombok. |
 | `-nna, --nonnull-annotations <arg>`                    | Adds a list of non-null annotations separated by a comma to be acknowledged by Annotator (e.g., com.example1.Nonnull,com.example2.Nonnull) |
 | `eic, enable-impact-cache`                             | Enables fixes impacts caching for next cycles. |
-| `-ll, --language-level <arg>`                          | Java language level used by the parser when reading source files. Supported values: `11`, `17`, `21`. Defaults to `17`. Use `21` when the target project uses Java 21 features such as record patterns or switch pattern matching. |
+| `-ll, --language-level <arg>`                          | Java language level used by the parser when reading source files. Accepts any release the bundled parser supports (`11` … `26`). Defaults to `17`. Set it to the release the target project compiles with, otherwise its syntax fails to parse — e.g. record and switch patterns need `21`, unnamed variables (`_`) need `22`. |
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ Here are some useful __optional__ arguments that can alter the default configura
 |--------------------------------------------------------|-------------|
 | `-n,--nullable <arg>`                                  | Sets custom `@Nullable` annotation. |
 | `-rboserr, --redirect-build-output-stderr`             | Redirects build outputs to `STD Err`. |
-| `-ll, --language-level <arg>`                          | Parser language level. Supported values: `11`, `17`, `21`. Defaults to `17`. Required when the target project uses Java 21 syntax (e.g. record patterns, switch patterns). |
+| `-ll, --language-level <arg>`                          | Parser language level. Accepts any release the bundled parser supports (`11` … `26`). Defaults to `17`. Set it to the release the target project compiles with, otherwise its syntax fails to parse. |
 
 
 To learn more about all the __optional__ arguments, please refer to [OPTIONS.md](./OPTIONS.md)
--- a/annotator-core/build.gradle
+++ b/annotator-core/build.gradle
@@ -98,6 +98,15 @@ if (!JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_21)) {
     }
 }
 
+// Java25Test builds a target project with language features released after Java 21.
+if (!JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_25)) {
+    test {
+        filter {
+            excludeTestsMatching "edu.ucr.cs.riple.core.Java25Test"
+        }
+    }
+}
+
 // temporarily exclude LombokTest until we find a solution for the issue.
 test {
     filter {
--- a/annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
+++ b/annotator-core/src/main/java/edu/ucr/cs/riple/core/Config.java
@@ -25,6 +25,7 @@
 package edu.ucr.cs.riple.core;
 
 import com.github.javaparser.ParserConfiguration;
+import com.google.common.base.Enums;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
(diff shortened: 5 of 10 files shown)


msridhar (approved):
Looks great, thanks!

msridhar (changes requested):
@AdzerKI would it be possible to add a unit test for the Java 25 support?  If it's tricky to figure out, let us know and we can help

AdzerKI:
Added in `ConfigurationTest`:

- `testLanguageLevelFlag` — the default (`17`), `21`, `25` and `17_PREVIEW` map to the matching `LanguageLevel`, and an unknown value still fails with `Unsupported language level: ...`.
- `testLanguageLevelSurvivesSerialization` — writes and reads back every `LanguageLevel` constant through `Config.Builder.write` / `Config(Path)`.

The round-trip test caught an existing asymmetry: `write` serialized the level as `name().split("_")[1]`, so `JAVA_1_0` came back as `1` (rejected on read) and `JAVA_17_PREVIEW` silently degraded to `JAVA_17`. It now strips only the `JAVA_` prefix, the exact inverse of parsing. That path was harmless while only `11`, `17` and `21` were accepted, but this PR opens it to preview levels.

I did not add an integration test in the shape of `Java21Test`: it drives a real Error Prone build of the target project, which caps the usable release at what the bundled `error_prone_core` supports, not at what the parser accepts. The flag change is entirely in `Config`, so unit-level coverage is what actually exercises it.

The earlier CI failure was `spotlessJavaCheck` on my formatting — `spotlessApply` is included in this push and the build is green locally.

msridhar:
I'd like an integration test showing we don't crash on code using Java 22+ features if possible, please.  Thanks!

msridhar:
@AdzerKI any updates on adding an integration test here?  If you need help or can't figure it out let us know.  Thanks!

AdzerKI:
@msridhar integration test added: `Java22Test`, following the `java-17`/`java-21` pattern.

It runs the full pipeline against a target project whose source uses an unnamed variable (`for (String _ : items)`, JEP 456, released in Java 22): the level is resolved through the bundled parser, the file is parsed and the expected `OnField` fix is reported. On the old `-ll` switch that project aborted with `Unsupported language level: 22` before inference started, which is exactly issue #240.

Mechanics kept the same as the existing levels: template `templates/java-22`, a `parser-configuration-22` CI job on JDK 22, and the test filtered out on JDKs that cannot compile the target project (`build` on JDK 11 stays unaffected). I stayed on 22 rather than a newer release because the Gradle wrapper here is 8.9, which does not run on JDK 25 (`Unsupported class file major version 69`); once the wrapper is bumped the same test template can move up.

Verified locally on Temurin 22: `:annotator-core:test --tests "*Java22Test*"` green, `spotlessCheck` clean.

.github/workflows/continuous-integration.yml
```diff
@@ -112,4 +112,18 @@ jobs:
...
+  parser-configuration-22:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 22
+        uses: actions/setup-java@v4
+        with:
+          java-version: 22
+          distribution: 'temurin'
+      - name: Build with Gradle
```
msridhar, line 127: Let's use JDK 25 here, since it's an LTS version, if that's ok?

annotator-core/src/test/java/edu/ucr/cs/riple/core/Java22Test.java
```diff
@@ -0,0 +1,63 @@
...
+import edu.ucr.cs.riple.core.tools.TReport;
+import edu.ucr.cs.riple.injector.location.OnField;
+import java.util.Set;
+import org.junit.Test;
+
+/**
+ * Tests in this class are related to Java 22 features. These tests include blocks of code that are
+ * not syntactically supported by Java 21.
+ */
+public class Java22Test extends AnnotatorBaseCoreTest {
```
msridhar: Similar to above, maybe let's just do `Java25Test` here

msridhar (changes requested):
Looks great!  Just a couple of minor comments

.github/workflows/continuous-integration.yml
AdzerKI: JDK 25 needs a newer Gradle than the wrapper carries. `gradle-8.9` runs on Java up to 22 (23 needs 8.10, 24 needs 8.14, 25 needs 9.1.0), and Gradle 9 itself requires Java 17+ to run, so moving the wrapper would raise the floor for the six JDK 11 jobs and for `sourceCompatibility = 1.11` in `build.gradle`. The integration test drives a real Gradle build of the target project with that same wrapper (`AnnotatorBaseCoreTest` copies `gradlew` into the temp project), so the target compile is capped identically; `templates/java-22/build.gradle` also pins `error_prone_core:2.31.0`, released a year before JDK 25, which would need its own bump.

22 is the lowest release that reproduces #240 (unnamed variables, JEP 456), so this job covers the regression as it stands. If you want the test on an LTS, I am happy to open a separate PR for the toolchain move (wrapper -> 9.x, JDK floor 11 -> 17, Error Prone in the templates) and rebase this one on top of it - just say which way you prefer.

annotator-core/src/test/java/edu/ucr/cs/riple/core/Java22Test.java
AdzerKI: Name follows the JDK the job actually runs, which is capped at 22 by the Gradle wrapper (details in the CI thread). It renames to `Java25Test` together with the toolchain bump, whenever that lands.

.github/workflows/continuous-integration.yml
msridhar: @AdzerKI good point!  If you could open a separate PR as you describe (wrapper -> 9.x, JDK floor 11 -> 17), I could approve that, and then we could rebase this one on top.  NullAway now requires JDK 17 anyway.  Thanks a lot!

AdzerKI: Opened as #280: wrapper 8.9 → 9.7.1, JDK floor 11 → 17, plus the plugin bumps Gradle 9 forces and the `--add-exports` the Error Prone test helper needs from JDK 16 on. Green on Temurin 17, and `Java21Test` still green on Temurin 21.

Once it lands I will rebase this one on top and move the integration test and its CI job to JDK 25, renaming `Java22Test` to `Java25Test` with it.

AdzerKI:
@msridhar rebased on master now that #280 is in.

Moving this test and its job to JDK 25 needs one more step first, so I opened #281: Error Prone 2.14.0 dies on JDK 25 inside `ASTHelpers` (`TypeTag.UNKNOWN` is gone from the JDK), and the target-project templates pin 2.31.0 and fail identically, so `:Target:compileJava` never gets to run. That is what keeps this PR on 22 rather than an LTS.

#281 takes Error Prone to 2.42.0 — the last release that still loads on the JDK 17 floor #280 set and already handles JDK 25 — plus the NullAway bump it forces and a one-line fix where NullAway's quoted field names leaked into the parse. Green on 17 and 21.

Once it lands I will rebase this one on top and switch the integration test and its CI job to JDK 25, renaming `Java22Test` to `Java25Test`.

msridhar:
@AdzerKI #281 is now merged, so hopefully we can finish this one up!

AdzerKI:
@msridhar rebased on master and moved the integration test to JDK 25: `Java25Test` with the `java-25` template and a `parser-configuration-25` job. It covers two constructs Java 21 cannot parse — a module import declaration (JEP 511, final in 25) and an unnamed variable (JEP 456, the repro from #240) — both driven through the full pipeline at `-ll 25`.

One thing to flag about the earlier run here: `build-without-cache` failed in `DeepTest`, `AnalysisModeTest` and `CoreTest` with

```
java.lang.RuntimeException at AnnotatorBaseCoreTest.java:88
    Caused by: java.io.FileNotFoundException at FileUtils.java:2719
```

That is `setup` copying the whole repository — `build/`, `.gradle/` and `.git/` included — while Gradle writes into them, from four forks at once. It is the same failure that took `build-with-gradle-build-action-cache` on master at 63074f6, not something this branch introduces. #282 has the fix: the copy skips those three directories.

AdzerKI:
The red job here is the copy race, not this PR: `CoreTest.multipleReturnNullable` died in `FileUtils.copyDirectory` with a `FileNotFoundException` while copying the repository into the test scratch dir — a file under a build directory vanished mid-copy. That is what #282 fixes by excluding the volatile directories from that copy. The other eight jobs, including `parser-configuration-25`, are green.

I cannot rerun jobs here — please either rerun that one or merge #282, and I will rebase.

msridhar (approved):

msridhar:
This looks great now @AdzerKI ! Thanks so much for the contribution!

PR #280, opened 2026-08-19
Build with Gradle 9 on JDK 17

AdzerKI:
## Problem

The build runs on Gradle 8.9 and targets Java 11. That caps the JDK the build itself can run on — `gradle-8.9` supports Java up to 22 — so an integration test for a language level newer than that cannot run at all, which is why #279 stopped at Java 22 instead of an LTS. NullAway already requires JDK 17, and Error Prone dropped Java 8 long ago.

Per @msridhar's request in #279: wrapper to 9.x, JDK floor 11 to 17, so #279 can rebase on top and move its integration test to JDK 25.

## Changes

- Wrapper 8.9 → 9.7.1 (`./gradlew wrapper`, so the scripts and the jar move with it).
- Java floor 11 → 17: `sourceCompatibility`/`targetCompatibility` and every CI job.
- Plugins that do not work on Gradle 9 are bumped: `net.ltgt.errorprone` (root, `ban-mutable-static`, the test templates and the sample), `com.diffplug.spotless` (also renamed from `com.diffplug.gradle.spotless`), `com.vanniktech.maven.publish` — where `publishToMavenCentral` no longer takes a `SonatypeHost` — and Shadow, which moved to `com.gradleup.shadow` with `from components.shadow` replacing `project.shadow.component(publication)`.
- Dropped what only existed for Java 8: the `errorproneJavac` dependency and the Error Prone version fallback in `gradle/dependencies.gradle`.
- Gradle 9 removals: `archivesBaseName` → `base { archivesName }`, and the project-level `sourceCompatibility`/`targetCompatibility` moved into the `java` extension.
- Test JVMs get the `--add-exports`/`--add-opens` for `jdk.compiler`. `CompilationTestHelper` in `annotator-scanner` and `checks/ban-mutable-static` reaches javac internals, which are open on JDK 11 and encapsulated from JDK 16 on; without the flags those 19 tests fail with `IllegalAccessError`.
- The Java-11 exclusion block is replaced by the guard the newer tests already use. `Java17Test` now runs everywhere, `Java21Test` runs on JDK 21+ (it builds a Java 21 target project), and `LombokTest` keeps its exclusion: it was excluded on Java 11, so with `build` running on Java 11 it never ran in CI, and on JDK 17 it still fails for the reason its comment mentions — unrelated to the toolchain.

## Verification

`./gradlew build` on Temurin 17: green, including `annotator-scanner` and `ban-mutable-static`, which had no CI coverage of their own before.
`./gradlew :annotator-core:test --tests "edu.ucr.cs.riple.core.Java21Test"` on Temurin 21: green — the `parser-configuration-21` job.

## Not in this PR

Gradle 9 warns that resolving `deps` from the root project through implicit parent lookup will fail in Gradle 10. Fixing it touches the dependency block of every module, so it seemed better as its own change.


annotator-core/src/test/resources/templates/java-21/build.gradle
```diff
@@ -60,7 +60,6 @@ subprojects {
         compileOnly "org.jspecify:jspecify:0.3.0"
         compileOnly "com.google.code.findbugs:jsr305:3.0.2"
         errorprone "com.google.errorprone:error_prone_core:2.31.0"
```
msridhar, line 62: On JDK 21, we should use the latest Error Prone (2.50.0).  But we don't need to do that in this PR.

gradle/dependencies.gradle
```diff
@@ -26,13 +26,10 @@
 def oldestErrorProneApi = "2.4.0"
 // Latest released Error Prone version that we've tested with
 def latestErrorProneVersion = "2.14.0"
```
msridhar, line 28: These versions are super old; we should update them (again, not for this PR).

msridhar (approved):
Looks great, assuming CI passes!

PR #281, opened 2026-08-22
Run the analysis stack on current Error Prone and NullAway

AdzerKI:
Follow-up to #280, and the toolchain half of #279.

## Why

The build cannot run on a current JDK, and the failure is in Error Prone, not in this project's code.

`error_prone_core:2.14.0` — the version the repo compiles itself with — dies on JDK 25 the moment a check touches the AST:

```
error: An unhandled exception was thrown by the Error Prone static analysis plugin.
     error-prone version: 2.14.0
     java.lang.NoSuchFieldError: Class com.sun.tools.javac.code.TypeTag does not have member field 'com.sun.tools.javac.code.TypeTag UNKNOWN'
  	at com.google.errorprone.util.ASTHelpers.<clinit>(ASTHelpers.java:1238)
```

The target-project templates hit the same wall one level down: they pin `2.31.0`, so `:Target:compileJava` inside the integration tests fails with the identical stack. That is what caps CI at JDK 22 today and blocks the JDK 25 job asked for in #279.

## What changes

- `latestErrorProneVersion` 2.14.0 -> 2.42.0, and the four target-project templates take that same version from `ERROR_PRONE_TEST_VERSION` instead of pinning `2.31.0` in four places. `annotator-core` exports it from `deps.versions.errorProne` next to `NULLAWAY_TEST_VERSION` and `ANNOTATOR_VERSION`.

  2.42.0 rather than the newest release, because 2.45.0 raised the runtime floor: from that version `error_prone_core` ships class file 65, so it cannot load on the JDK 17 that #280 established as this repo's floor —

  ```
  java.lang.UnsupportedClassVersionError: com/google/errorprone/ErrorProneJavacPlugin has been compiled by a more recent version of the Java Runtime (class file version 65.0), this version of the Java Runtime only recognizes class file versions up to 61.0
  ```

  2.42.0 is the last release that still runs on JDK 17 and already handles JDK 25 — checked on a JDK 25 build of a target project using an unnamed variable, which is the syntax #279 needs.

  If you would rather be on the current Error Prone, the change that gets you there without touching what the artifacts require of users is to split the build JDK from the release target: run the jobs on 21 and compile with `--release 17`, leaving `targetCompatibility` where it is. Say the word and I will do that in this PR instead.

  `oldestErrorProneApi` is untouched at 2.4.0, so `annotator-scanner` and the custom checks still build against the old API and the supported-runtime floor is unchanged.

- NullAway 0.10.19 / 0.12.4 -> 0.13.8. The repo carried two NullAway versions in two files (`versions.nullaway` for `library-model-loader`, `NULLAWAY_TEST` for the integration tests); `NULLAWAY_TEST` now reads `deps.versions.nullaway`, so there is one version.

- `extractUninitializedFieldNames` accepts NullAway's quoted syntax element references. Since 0.13.8 (#1620) the initializer message reads `'f' (line 3)` where it used to read `f (line 3)`, and the quotes ended up inside the field name: the registry lookup missed, `OnField{variables=['f']}` reports appeared next to the real ones, and ten `CoreTest`/`DeepTest` cases failed. The method's own comment asked for exactly this ("should be watched carefully and updated if the format is changed by NullAway"); it now tolerates both spellings, so the parse does not pin a NullAway version.

NullAway 0.14.0 is deliberately not taken: it writes serialization version 4 while `NullAway.VERSION` here is 3, so every integration test stops at `verifyCheckerCompatibility` with *"This Annotator version only supports NullAway serialization version 3, but found: 4"*. Reading version 4 is separate work; the version note in `dependencies.gradle` records the ceiling.

## Verified

`:annotator-core:spotlessCheck` and `:annotator-core:test` green on Temurin 17 and 21, and `:checks:ban-mutable-static:compileJava` plus `:annotator-core:compileJava` green on 17, 21 and 25.

With this in, #279 rebases on top and moves its integration test and CI job to JDK 25.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -24,8 +24,9 @@
 
 // The oldest version of Error Prone that we support running on
 def oldestErrorProneApi = "2.4.0"
-// Latest released Error Prone version that we've tested with
-def latestErrorProneVersion = "2.14.0"
+// Latest Error Prone that still runs on the JDK 17 floor, also used by the test target projects.
+// 2.45.0 and later ship class file 65 and need JDK 21 to load.
+def latestErrorProneVersion = "2.42.0"
 
 
 def versions = [
@@ -39,7 +40,8 @@ def versions = [
         commonsio               : "2.11.0",
         progressbar             : "0.9.2",
         junitjupiter            : "5.7.2",
-        nullaway                : "0.10.19",
+        // Latest NullAway that still writes the serialization version NullAway.VERSION reads
+        nullaway                : "0.13.8",
         mockito                 : "5.2.0",
         junit                   : "4.13.2"
 ]
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -23,7 +23,7 @@
  */
 
 // The oldest version of Error Prone that we support running on
-def oldestErrorProneApi = "2.4.0"
+def oldestErrorProneApi = "2.36.0"
 // Latest Error Prone that still runs on the JDK 17 floor, also used by the test target projects.
 // 2.45.0 and later ship class file 65 and need JDK 21 to load.
 def latestErrorProneVersion = "2.42.0"
--- a/annotator-core/build.gradle
+++ b/annotator-core/build.gradle
@@ -54,14 +54,12 @@ spotless {
     }
 }
 
-// Should be the latest supporting version of NullAway.
-def NULLAWAY_TEST = "0.12.4"
-
 tasks.test.dependsOn(':annotator-scanner:publishToMavenLocal')
 
 // Set up environment variables for test configuration tu run with the latest development version.
 tasks.test.doFirst {
-    environment "NULLAWAY_TEST_VERSION", NULLAWAY_TEST
+    environment "NULLAWAY_TEST_VERSION", deps.versions.nullaway
+    environment "ERROR_PRONE_TEST_VERSION", deps.versions.errorProne
     environment "ANNOTATOR_VERSION", project.version
 }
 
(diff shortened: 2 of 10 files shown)


gradle/dependencies.gradle
```diff
@@ -24,8 +24,9 @@
 
 // The oldest version of Error Prone that we support running on
 def oldestErrorProneApi = "2.4.0"
```
msridhar: Maybe bump this to 2.36.0, oldest version supported by NullAway?

https://github.com/uber/NullAway/blob/cf7c47dd0673d076d4c7fe4dcb7fcd3a4f1e87f4/gradle/libs.versions.toml#L20

gradle/dependencies.gradle
```diff
@@ -24,8 +24,9 @@
...
-// Latest released Error Prone version that we've tested with
-def latestErrorProneVersion = "2.14.0"
+// Latest Error Prone that still runs on the JDK 17 floor, also used by the test target projects.
+// 2.45.0 and later ship class file 65 and need JDK 21 to load.
+def latestErrorProneVersion = "2.42.0"
```
msridhar, line 29: 👍 This matches NullAway:

https://github.com/uber/NullAway/blob/cf7c47dd0673d076d4c7fe4dcb7fcd3a4f1e87f4/gradle/libs.versions.toml#L19

msridhar (commented):
looks great!  one minor comment

msridhar (approved):

msridhar:
@AdzerKI went ahead and fixed a few remaining issues here.  Thanks for the contribution!

PR #282, opened 2026-08-23
Skip volatile directories when copying the repository into a test scratch dir

AdzerKI:
## Problem

`AnnotatorBaseCoreTest.setup` copies the whole repository into the test's scratch dir:

```java
FileUtils.copyDirectory(repositoryDirectory.toFile(), outDirPath.resolve("Annotator").toFile());
```

That walk includes `build/`, `.gradle/` and `.git/`, which the outer Gradle build keeps writing into while the tests run, and four test forks walk them at once. When a file disappears between `list()` and opening it the copy throws and the test dies before it starts:

```
DeepTest > paramPassTest FAILED
    java.lang.RuntimeException at AnnotatorBaseCoreTest.java:88
        Caused by: java.io.FileNotFoundException at FileUtils.java:2719
```

It is a race, so it lands on whichever tests happen to be running: `build-without-cache` took `DeepTest`, `AnalysisModeTest` and `CoreTest` on one run of #279, and `build-with-gradle-build-action-cache` failed the same way on master at 63074f6.

## Changes

The copy skips `build`, `.git` and `.gradle`. The copy exists to run `./gradlew library-model-loader:jar --rerun-tasks` against the sources, which rebuilds what it needs, so none of the three is read from it. It also stops copying the build outputs of the whole repository into a temp dir once per test.

## Verification

`DeepTest.paramPassTest`, `CoreTest.multipleReturnNullable`, `AnalysisModeTest` and `Java21Test` green on JDK 21; `spotlessCheck` clean.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -53,10 +53,13 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew build --scan --no-daemon
       - name: Archive test report
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: Test report
-          path: build/reports/tests/test
+          path: |
+            **/build/reports/tests/test
+            **/build/test-results/test
   parallel-processing-enabled-caching-disabled:
     runs-on: ubuntu-latest
     steps:
--- a/annotator-core/src/test/java/edu/ucr/cs/riple/core/AnnotatorBaseCoreTest.java
+++ b/annotator-core/src/test/java/edu/ucr/cs/riple/core/AnnotatorBaseCoreTest.java
@@ -26,9 +26,11 @@ package edu.ucr.cs.riple.core;
 
 import edu.ucr.cs.riple.core.tools.CoreTestHelper;
 import edu.ucr.cs.riple.core.tools.Utility;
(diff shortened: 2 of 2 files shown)


AdzerKI:
Correction on my own claim above: the copy filter alone does not stop the failure. The run on this branch still lost `CoreTest > multipleReturnNullable` to the same `FileNotFoundException`, so at least one source of the copy still changes under it, and it is not `build`, `.git` or `.gradle`.

What the filter does do is worth keeping on its own: it stops copying the build outputs and the VCS metadata of the whole repository into a scratch dir 59 times, once per test, for a copy that only feeds `./gradlew library-model-loader:jar --rerun-tasks`.

To find the rest I added a second commit to this branch: `Archive test report` now runs on failure too, and its path points at where Gradle actually writes the reports. Today it is `build/reports/tests/test` — the root project, which has no test task — so the artifact has been empty on every green run and absent on every red one. With the reports and the XML results collected per subproject, the next occurrence names the file that vanished instead of just the exception type.

I did not reproduce it locally: the whole `annotator-core` suite is green here both serially and with four forks (JDK 17, 59 tests).

msridhar (approved):

